PR #36, opened 2016-11-18
Find ASGs by DescribeTags first since DescribeAutoScalingGroups is paginated

jwineinger:
Our environment has more than one page of autoscaling groups, so just iterating over a single page of results was missing the ASG I was testing with. This PR uses DescribeTags with a filter to find the ASG names that have the tag and then do a DescribeAutoScalingGroups only on those groups.

### DIFF
--- a/core/region.go
+++ b/core/region.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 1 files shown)


cristim:
Thanks, that's a really good catch, I never tested it on an account with so many groups.

I guess this will still remain a problem if you ever enable it on all your groups, so eventually we'd need to add pagination to most of our API calls.

Issue #25, opened 2016-10-29
Consider availability zone when picking an instance type


nmeierpolys:
Our instances all run in the us-east-1d availability zone. When the autospotting process runs, it's finding the cg1.4xlarge instance type as the cheapest option and trying to use that to request spot instances. Unfortunately, the cg1.4xlarge instance type isn't available in that us-east-1d AZ, only us-east-1c. We get this error on the spot request "capacity-not-available: There is no Spot capacity available that matches your request. "

For use cases like ours that are limited to a specific AZ, it would be really helpful to consider the AZ when retrieving spot pricing info, and only pick an instance type if it's available in the AZ. 

When this happens, it continues to request a new spot instance each time the process runs, which fairly quickly uses up the maximum number of open spot instance requests that AWS allows, preventing other spot instance requests.

nmeierpolys:
Also, when this happens, it continues to request a new spot instance each time the process runs, which fairly quickly uses up the maximum number of open spot instance requests that AWS allows, preventing other spot instance requests.

nmeierpolys:
I was able to get past this for our instances by changing the code to skip that instance type and hosting our own binaries. I'll see if I can figure out why it's not handling the case better in general and hopefully put in a PR when I get a chance.

cristim:
Good catch @nmeierpolys, thanks for reporting this!

Off-topic: I'd like to learn more about your use case for only running in a single region, for reliability concerns that's not something I would normally do, especially with spot instances, and I'm always curious to learn about such interesting use cases happening 'in the wild'.

Regarding the issue you noticed, that sounds like a real problem. As far as I can see in the spot instance pricing history in my AWS console, that instance type is indeed only available in a single AZ in US-East-1(in my account that is labeled as us-east-1a), so I guess the algorithm is somehow not checking against this edge case.

I will try to reproduce this in my own AWS account and I will have a look at the source code to see why does that unavailable instance type appear to be the cheapest.

Somehow related to this issue, I think I could implement a way to track instance launch failures for a given instance type/AZ combination, and somehow temporarily blacklist them for a few hours if the instance fails to launch over multiple autospotting runs in a row. I will create a new issue for implementing this kind of feature.

nmeierpolys:
Thanks @cristim. Right now, we're running in a single AZ because our system and deploys expect everything to be in a single subnet. There's nothing technically preventing us from changing this, it just hasn't been a big enough priority to commit time to it yet.

[...]

PR #46, opened 2016-11-21
Big refactoring

cristim:


cristim:
@roeyazroel, @nmeierpolys, @jwineinger: please try this out and help me get this code reviewed.

This is mostly just a code cleanup, there shouldn't be any functionality changes.

Any suggestions about writing this in a more idiomatic go that allows me to easier write automated tests for it are more than welcome.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+	logger.Printf("\n Found following compatible instances: %#v\n",
+		filteredInstanceTypes)
+	return filteredInstanceTypes, nil
+
 }
 
-func (it *instance) isSpot() bool {
-	return (it.InstanceLifecycle != nil &&
-		*it.InstanceLifecycle == "spot")
+func (i *instance) getCheapestCompatibleSpotInstanceType() (*string, error) {
```
xlr-8: Isn't it in this method that you are supposed to check that the SpotInstance that you get is in the same AZ as the ASG running? Ref: https://github.com/cristim/autospotting/issues/25

core/launch_configuration.go
```diff
@@ -0,0 +1,143 @@
...
+	logger.Printf("Launch configuration would attach %d ephemeral volumes",
+		"if available", count)
+
+	return count, nil
+}
+
+func (lc *launchConfiguration) convertLaunchConfigurationToSpotSpecification(
+	baseInstance *instance,
+	instanceType string,
+	az string) *ec2.RequestSpotLaunchSpecification {
```
xlr-8: Here we do take the AZ parameter, but as stated earlier we don't check that the compatible instance, is also in the AZ of the ASG.

core/region.go
```diff
@@ -246,14 +249,13 @@ func (r *region) scanForEnabledAutoScalingGroupsByTag(asgs *[]*string) {
...
-		return
 	}
+	return asgs
 }
 
 func (r *region) scanForEnabledAutoScalingGroups() {
-	asgNames := []*string{}
 
-	r.scanForEnabledAutoScalingGroupsByTag(&asgNames)
+	asgNames := r.scanForEnabledAutoScalingGroupsByTag()
```
xlr-8: I would have renamed this method to something a bit different, than what this method is doing/named. Something like: `getEnabledAutoScalingGroups` for example, as the names of those methods is very similar but don't fulfil the same purpose, it's a bit confusing, and adding `ByTag` is a bit pointless, as it's currently the only way of enabling ASGs.

core/spot_instance_request.go
```diff
@@ -0,0 +1,78 @@
...
+
+	logger.Println(s.asg.name, "Done waiting for an instance.")
+
+	// Now we try to get the InstanceID of the instance we got
+	requestDetails, err := ec2Client.DescribeSpotInstanceRequests(&params)
+	if err != nil {
+		logger.Println(s.asg.name, "Failed to describe spot instance requests")
+	}
+
+	// due to the waiter we can now safely assume all this data is available
```
xlr-8: Except in the very unlikely event of someone deleting the request in the meantime?

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+			continue
+		}
+
+		// checking how many spot instances of this type we already have, so that
+		// we can see how risky it is to launch a new one.
+		spotInstanceCount := i.asg.alreadyRunningSpotInstanceCount(
+			candidate.instanceType, availabilityZone)
+
+		// We skip it in case we have more than 20% instances of this type already
+		// running
```
cristim: Yes, we should raise an issue about this, it's indeed quite cryptic at the moment.

20% per AZ was just a starting point that would spread the instances enough to give some safety in case of termination, but now we should make it configurable.   

I can't reply directly to the first message about the logger, so here it comes:

To be honest my golang skills are still relatively rudimentary, the way I did the logger is definitely problematic in tests, I'm open for suggestions, but let's do it in another issue.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+		if spotInstanceCount == 0 ||
+			(*i.asg.DesiredCapacity/spotInstanceCount > 4) {
+			logger.Println(i.asg.name,
+				"no redundancy issues found for", candidate.instanceType,
+				"existing", spotInstanceCount,
+				"spot instances, adding for comparison",
+			)
+
+			filteredInstanceTypes = append(filteredInstanceTypes, candidate.instanceType)
```
cristim: Fair point, that would speed up things a bit and especially reduce the logging noise.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+				"are not compatible or resulting redundancy for the availability zone",
+				"would be dangerously low")
+
+		}
+
+	}
+	logger.Printf("\n Found following compatible instances: %#v\n",
+		filteredInstanceTypes)
+	return filteredInstanceTypes, nil
+
```
cristim: Yes, totally I agree this is not the best way to write this. I'll try to come up with a better way.

core/instance.go
cristim: I'm not sure if I understand your point, but the AZ of the spot instance is always taken from the on-demand instance we are replacing with it, which was for sure added by AutoScaling to one of the group's configured AZs. That's why we can't launch instances in different AZs.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+			chosenInstanceType, "priced at", minPrice)
+	}
+
+	if chosenInstanceType != "" {
+		logger.Println("Chose cheapest instance type", chosenInstanceType)
+		return &chosenInstanceType, nil
+	}
+	logger.Println("Couldn't find any cheaper spot instance type")
+	return nil, fmt.Errorf("No cheaper spot instance types could be found")
+
```
cristim: I agree, this would make more sense.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+	svc := i.region.services.ec2
+	params := ec2.CreateTagsInput{
+		Resources: []*string{i.InstanceId},
+		Tags:      tags,
+	}
+
+	logger.Println(i.region.name, "Tagging spot instance", *i.InstanceId)
+
+	for _, err := svc.CreateTags(&params); err != nil; _, err = svc.CreateTags(&params) {
```
cristim: Even though I wrote this not so long time ago, I also find it hard to understand. This code is a bit write-only and may need some refactoring :)

What is does is trying to tag the instance in a loop until the tagging returns no error, since the instance has just been launched and some API calls executed against it may be failing for a short while immediately after launch.

core/instance.go
```diff
@@ -18,53 +44,255 @@ type instanceTypeInformation struct {
...
+
+		logger.Println(i.region.name,
+			"Failed to create tags for the spot instance", *i.InstanceId, err.Error())
+
+		logger.Println(i.region.name,
+			"Sleeping for 5 seconds before retrying")
```
cristim: That would be much prettier than that cryptic one-liner of mine.

core/launch_configuration.go
cristim: Yes, because the compatible instance is always in the same AZ as the on-demand instance it is replacing, which can't be outside the group's AZs.

core/region.go
cristim: Fair point.

core/spot_instance_request.go
cristim: Why would someone do this?

But yes, I guess an explicit check wouldn't hurt.

core/instance.go
xlr-8: I'm also fairly new to golang, so no worries.

Why is it a problem in tests?
For the log tool, I don't have any preference I have heard of https://github.com/Sirupsen/logrus and https://github.com/golang/glog, but apparently there are many others such as: https://github.com/inconshreveable/log15 or https://github.com/op/go-logging.

Considering the fairly basic usage (having log levels), I think any could do, it's only a matter of preference and/or evolution afterwards.

core/instance.go
xlr-8: I miss understood something, which is why I edited it afterwards, no worries.

core/instance.go
xlr-8: What I have in mind would be:
```golang
iterate over spot_possibilities
   ! check number running + previously saved price
       continue
   ! check compatibility class
       continue
   ! check compatibility storage
       continue
   save as_most_interesting_offer 
```

Where `check XXX` are methods, checking those appropriately. We could also consider merging those check in one, to have only 1 check + 1 continue, and in each of those methods have the appropriate log debug/info.

core/instance.go
cristim: The problem I had was due to the fact that the logger can't be used from tests right away, without having it initialized before in the sest preparation step.

Thanks, I'll have a look at those logging libraries.

core/instance.go
xlr-8: OK, I understand better now; indeed it would be better to implement a retry mechanism with a timeout. Isn't there extra things to check before tagging? such as: make sure the instance is running, or all info are fulfilled? I know the tagging action and starting action are 2 separated things, but I thought, maybe we could have a smarter loop/call here.

core/launch_configuration.go
xlr-8: Yup, which makes the issue about the AZ quite weird. I'm wondering if this is not the data file, that might be incorrect.

core/spot_instance_request.go
xlr-8: Because people. :)
I was just pointing, that we can't be sure that all data is "safely" available, since someone or another program could have make a call in between.
Considering the unlikelihood of this happening, I honestly don't think it's worth adding a check, I make the same assumption in the 'configurable algorithm': where I don't double check the tags between 2 methods.

core/instance.go
cristim: Looks great, please try to make a pull request with this and the other suggestions in this review when you have some time.

core/launch_configuration.go
cristim: I'm not sure what was the issue there, it definitely needs some more investigation.

core/spot_instance_request.go
cristim: I agree.

If the user does something that is interfering with the logic then we can't guarantee the result and we'd need to be much more defensive in our checks.

Let's only fix it if it turns out to be a problem in the wild.

core/instance.go
cristim: I raised #59, #60 and #61 about this.

Issue #47, opened 2016-11-23
Whitelist/blacklist certain instance types via configuration


cristim:
I saw forks that hard-coded some specific instance types they need in their environment. Other people complained that some instances are problematic in certain availability zones and had them hardcoded out.

This could be configurable using CloudFormation stack parameters, passed as variables to the Lambda function now that Lambda supports this feature. We could add two new CloudFormation stack options:

```
WhiteListedInstanceTypes: m3.medium,c4.large
BlackListedInstanceTypes: c3.large
```

On another hand I think that such a global setting may not be desirable for some cases, so it may be better to also be able to apply it on a per-group basis, using additional tags set on the AutoScaling group:

```
AutoSpotting_WhiteListedInstanceTypes: c3.large,c4.large
AutoSpotting_BlackListedInstanceTypes:  c3.xlarge
```

roeyazroel:
I think that it will be very useful.

Maybe also add "family" tag?  
AutoSpotting_WhiteListedInstanceTypesFamily: c3, c4
AutoSpotting_BlackListedInstanceTypesFamily:  m4

cristim:
@roeyazroel I think it's better to have it based on globs, just like we have for the new region filter.

As for the names of the configuration options, they should be underscore_separated_only, just like we have the new options used for the on-demand capacity configuration.

cristim:
This is being implemented in #120

cristim:
#120 implements instance type whitelisting without globs as well as an option of keeping the original instance type, whatever that is.

cristim:
I think we can close this, recent builds support both filtering modes

xlr-8:
Has the whitelisting been validated as bug less? I haven't got the chance to do so, some users seem to use it fine, while I remember some other reported issues.

cristim:
The reports I've seen so far were about side effects of running the replacement algorithm with fewer instance types to choose from.

Issue #21, opened 2016-09-29
Keep at least one (or more) reserved instance running


seme1:
I have a web application and it is essential to have a number of instances always running to handle the possibility of having all spot instances shut down and make the web application still accessible. 

Is it possible to configure Autospotting to keep a specific (or a number of) reserved instances without replacement while replacing the rest of the instances with spot ones ?

cristim:
Thanks for reporting this.

It looks like a duplicate of #10, but we can discuss this here, because we have better context in the content of the issue.

Honestly I think that even a small number of on-demand(and even reserved) instances may cost you as much as the entire rest of the group running on spot instances, and in case all those spot instances are terminated, the remaining reserved/on-demand capacity is likely to be insufficient for many use cases, at least for a short while.

AutoSpotting already does a lot to reduce the risk of termination for all your spot instances, by spreading them over multiple instance types on each availability zone when it can get dangerous, but I've seen this question being asked a lot so I will have to implement it somehow, and it's actually the next thing I plan to do.

I realize that there are so many use cases out there and if possible it's better to accommodate more of them and let the users decide how to use it. For that, I need to make autospotting configurable, so that everyone can enable/disable such optional functionality and tweak it as per their needs. This is already tracked as #5, we can discuss that aspect there if you would like. But once I make it configurable, I can relatively easily implement this functionality.

schodemeiss:
+1 for this functionality. Just a min configurable for the amount of EC2's to leave in on-demand would be great.

seme1:
@cristim Thanks for the response. I'd rather be ready for the worst case scenario (when all spot instances are terminated) by having a slow service (or with lots of timeouts) than having a completely non-functioning/offline service.

cristim:
@seme1 ideally you should have a multi-region setup so that a single region failure would not be catastrophic, but I can understand your point.

As I said, I will implement this functionality soon, I just need to implement the configuration mechanism first, please bear with me.

In the meantime, if you really need this functionality now, it's actually pretty easy to fork autospotting and build your own version that implements this functionality in just a few lines of code. Have a look in core/autoscaling.go, around line 800 where we count the current number of spot instances.

You basically need exit that loop if that number is too close to the group's total capacity, it's a simple subtraction followed by an if statement. You can also make it percentage-wise with just a few more lines.

You will also need an S3 bucket where to deploy the forked builds, and deploy your CloudFormation stack based on that bucket.

All in all, shouldn't take more than a couple of hours of work.

[...]

Issue #12, opened 2016-08-19
Investigate weighting of instance types


cristim:
This may allow us to replace a number of smaller instances with a single bigger instance, as long as the price is proportionally lower.

xlr-8:
Could we discuss what would be the use case?

Considering that the algorithm currently tries to get the most powerful instance for the lowest price, I have trouble to think about when using it. Would it be to try to save some more, for something more powerful anyway?

cristim:
It's all about minimizing the price per desired unit of capacity.

All the prices mentioned below are fictional examples assumed to be stable for a relatively long period of time.

Let's say you run on c3.large on-demand instances and the application is CPU-intensive like many tend to be

- The c3.large is usually around $0.1/hour on the spot market but the current price is 0.5 and you need to replace that on-demand instance now.
- At this time the almost identical c4.large costs 0.18
- The c3.xlarge that has 2x capacity costs 0.2.

The algorithm would currently give you the less powerful c4.large instead of the c3.xlarge which costs just slightly more but is twice as powerful.

This feature would allow the user to divide the cost by the number of CPU cores of the candidate spot instance type so that would launch a larger instance that costs least per CPU core.

The algorithm would need to be configurable to allow the user to specify which unit of capacity is more important for the application (CPU, memory, maybe instance store disk, etc. but CPU and memory are good for a start) and based on that to select the instance type that optimizes for more capacity for the least cost.

As usual, that value should be passed as a tag set on each group, and when absent the algorithm should behave as it does now.

Issue #10, opened 2016-08-19
Allow keeping some residual on-demand capacity when configured as such


cristim:
This depends on #5, but once that's in place we can have a way to configure the algorithm to keep some on-demand capacity.

cristim:
Fixed in #54

PR #54, opened 2016-12-15
algo: Making the minimum on-demand value configurable

xlr-8:
The configuration regarding the minimum instance number can be
configured from the CLI as well as from the tag of the ASG.

The priority goes as following:
* Tag - Strict number
* Tag - Percentage
* CLI - Strict number
* CLI - Percentage

Which means the CLI parameters can be overidden by the tags, but
percentages value can never overide the strict numbers - except if
the value is invalid (out of range, negative, etc). Which means,
that as soon as a good value is found the processing stops.

The percentage is the percentage of instances currently in the ASG,
not Max/Min/Desired. Please note that the values are rounded up,
so use it with logic.

If the number of On-Demand is above the minimum required, then the
processing of the ASG continues.
If the value is equal, then nothing needs to be done, the processing
stops.
If the value is below, then it means that a spot-instance need to be
terminated in order for the ELB to recreate an On-Demand instance.

I believe that with the current way of making the algorithm
configurable, it shouldn't be too complicated to implement extra
elements.

Reviews are more than welcome.

Extra notes to consider:
*  ~~many if/else are for logging/debugging and could be removed~~
* tag values can obviously be changed, it was just to test the functioning
* added date/time in the logs to make it, more like proper logs
* some logs about regions have been moved to 'debug' to avoid extra unnecessary information
* should help and/or close such issues: https://github.com/cristim/autospotting/issues/5 https://github.com/cristim/autospotting/issues/10 https://github.com/cristim/autospotting/issues/12 https://github.com/cristim/autospotting/issues/21 https://github.com/cristim/autospotting/issues/47


cristim:
Thanks for this contribution, it's looking great at a first glance, I hope I'll have some time to review it in detail over the week-end.

xlr-8:
Tests are expected to come - this is why it's still a [WIP], however I'll be on holiday soon and don't know how much time I'll have dedicated for that, so it might have to wait a bit.

xlr-8:
All the new code has been covered, feel free to review!

core/autoscaling.go
```diff
@@ -20,16 +27,149 @@ type autoScalingGroup struct {
...
+func (a *autoScalingGroup) loadPercentageOnDemand(tagValue *string) (int64, bool) {
+	percentage, err := strconv.ParseFloat(*tagValue, 64)
+	if err != nil {
+		logger.Printf("Error with ParseFloat: %s\n", err.Error())
+	} else if percentage == 0 {
+		logger.Printf("Loaded MinOnDemand value to %f from tag %s\n", percentage, OnDemandPercentageLong)
+		return int64(percentage), true
+	} else if percentage > 0 && percentage <= 100 {
+		instanceNumber := float64(len(a.instances.catalog))
+		onDemand := int64(math.Floor((instanceNumber * percentage / 100.0) + .5))
```
cristim: I see this works out in the tests, but why add 0.5 here?

Shouldn't `math.Floor()` round up to the correct value without it?

autospotting.go
```diff
@@ -3,12 +3,11 @@ package main
 import (
 	"encoding/json"
 	"flag"
-	"fmt"
 	"log"
 	"os"
 
-	autospotting "github.com/cristim/autospotting/core"
 	lambda "github.com/eawsy/aws-lambda-go/service/lambda/runtime"
+	autospotting "github.com/cristim/autospotting/core"
```
cristim: I think our editors will clash here, we have to find a way to make them agree on the import order.

core/autoscaling.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"math"
+)
+
+const (
+	OnDemandPercentageLong  = "on_demand_percentage"
+	OnDemandNumberLong      = "on_demand_number"
```
cristim: I think we should prefix these with something, to make it more clear what they are about:

I am thinking at something like `autospotting_on_demand_percentage` and `autospotting_on_demand_number`

core/autoscaling.go
cristim: Never mind, I did the math and I understand it now

autospotting.go
xlr-8: For this, I do not know if it's a matter of editor or not, but I have a pre-commit hook checking `go fmt` on all modified files. So maybe it comes from here.

core/autoscaling.go
xlr-8: I'll comment just in case of someone else reading it. Basically, if you do no add a value (here 0.5), then Floor will always return 0, when the value is below 1. So in our tests with 3 instances 33% (which is in fact 0.333333333) will return 0.99 of an instance, and therefore Floor would return 0 instead of 1.

So the value 0.5 says, that if a percentage return more than 0.5 of an instance, to round that up to 1. I could add extra tests to make this more explicit, in a situation with 3 instances, 16% would be ~0.49 so would return 0 instance, but 17% or (16.66666667%) would return 1 instance, as it would be rounded up.

More information can be found here if necessary: https://github.com/golang/go/issues/4594

core/autoscaling.go
xlr-8: Sure, can be changed easily. Regarding the single tag combined, we could think about it, when adding more options, but for the 'on-demand' configuration where only one value has to be taken from whether percentage, whether number, it did not make that much sense to implement it.

Considering the number of tags allowed in ASG, I don't know if we will implement it, otherwise something like this could be a good base: https://play.golang.org/p/6Q8iV3paZx

xlr-8:
Too bad, you merged it before adding the documentation; I thought about it but forgot to write it down for some reasons. I'll try to write some documentation/explanations, whenever I get the chance.

core/autoscaling.go
cristim: That's fine, the combined mechanism was actually first proposed when we had a much smaller limit on tags, but in the meantime it was increased and now it is more than enough. I don't think it will become a problem anytime soon.

autospotting.go
cristim: I don't think it's go fmt, but it may be my use of goimports.

I will investigate this further.

cristim:
That's fine, I already saw your new pull request #56

xlr-8:
Yeah, but I also wanted to do some cleanups in the commits to have nicer history :) Will be for next time!

PR #55, opened 2016-12-20
Add support for the new eu-west-2(London) region, fixes #53

### LINKED ISSUE #53: Add support for the new London region

cristim:


cristim:
@xlr-8, @jwineinger: please review this.

cristim:
Unfortunately each local build will change the generated_bindata.go file, what we can do is try not to commit that unless we're doing a data update like this one.

I would prefer to have that ignored from git, like I had it initially, but unfortunately it is needed for 'go get'.

PR #56, opened 2016-12-22
doc: fixing tag names and adding documentation

xlr-8:
The on-demand values configuration is now available but was missing
some documentation/explanations on its functionning. That should
clarify things for new commers.

In order to make the documentation more clear and easy to browse a TOC
has also been added, using: https://github.com/ekalinin/github-markdown-toc


README.md
```diff
@@ -203,12 +231,67 @@ One good way to automate is using CloudFormation, using this example snippet:
...
 
-#### Elastic Beanstalk Installation ####
+#### For Elastic Beanstalk ####
 
 * In order to add tags to existing Elastic Beanstalk environment, you will
   need to rebuild the environment with the spot-enabled tag. Follow this
   [guide](http://www.boringgeek.com/add-or-update-tags-on-existing-elastic-beanstalk-environments)
 
+### Configuration of autospotting ###
+#### Default configuration ####
```
cristim: I would phrase this in such a way to make it clear that this is for manual execution, mainly for testing and evaluation, as compared to the execution from the Lambda runtime, which is actually the default operation mode. That's where it makes sense to be able to specify a certain region, in order to limit the impact when testing it. 

At the moment, when executed for Lambda we can't pass any of the options available as flags when executed locally, so it is processing all available regions and we can't pass the options for on-demand capacity, since none of that is  configurable for Lambda yet.

That could be implemented using CloudFormation stack parameters, which could be used to generate a certain custom CloudWatch event containing global options, that would need to be parsed and integrated with our current configuration mechanism.

xlr-8: Fair point, a maybe more common way would be to use a configuration file used by the program, but I'm not sure on how possible that would be within Lambda ?

cristim: Yes, I thought about that, but that would likely require a rebuild when changing the configuration, since the config file would need to embedded into the function code, as opposed to the CloudFormation mechanism which only requires some parameter changes.

The current implementation in #57 is based on information passed by the CloudWatch Event, but I just realized we could use the relatively new Lambda environment variables to achieve the same functionality. I'll give that a try since it may be cleaner than the CloudWatch Event approach and would also work for local execution.

PR #58, opened 2016-12-23
Make the function configurable from CloudFormation, fixes #57

### LINKED ISSUE #57: Implement global configuration options when executed from Lambda
Currently the autospotting CLI tool supports a number of flags when executed manually, but those are completely ignored when executed from Lambda, which is not yet configurable.

This should be implemented by customizing our CloudWatch events.
- The CloudFormation stack should have configuration parameters corresponding to each of the global command-line flags
- A custom CloudWatch event should be generated by the CloudFormation template based on those stack parameters
- The event data should be handled by the autospotting Lambda handler for generating configuration data structure just like done for the command-line flags

cristim:


cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -26,6 +31,21 @@
...
+      "Type": "Number"
+    },
+    "MinOnDemandPercentage": {
+      "Default": "0",
+      "Description": "Minimum on-demand instances (percentage) running in ASG.",
+      "Type": "Number"
+    },
+    "Regions": {
+      "Default": "",
+      "Description": "Regions (comma separated list)where it should run, by default runs on all regions",
```
xlr-8: typo with spacing between `list)where` I think it's also in the CLI option.

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -26,6 +31,21 @@
...
       "Description": "Number of days to keep the Lambda function logs in CloudWatch.",
       "Type": "Number"
+    },
+    "MinOnDemandNumber": {
+      "Default": "0",
+      "Description": "Minimum on-demand instances (number) running in ASG.",
+      "Type": "Number"
+    },
+    "MinOnDemandPercentage": {
+      "Default": "0",
```
xlr-8: Might use 0.0 to explicit the fact that is a float rather than an int?

cloudformation/stacks/AutoSpotting/parameters.yaml
```diff
@@ -1,3 +1,8 @@
 ---
 LambdaHandlerFunction: handler.handle
 LambdaS3Bucket: cloudprowess
+LambdaMemorySize: '256'
+ExecutionFrequency: "rate(1 minute)"
+MinOnDemandNumber: '1'
+MinOnDemandPercentage: '50'
+Regions: us-east-1,eu-west-1,ca-central-1
```
xlr-8: We currently have no checks for regions, while it would be used at runtime. It might be better to whether make a call to `describe-regions` to have something dynamic, or use `const` with a list of current region - but require to keep it updated. I'd be more in favour of the API call, as it wouldn't require commits whenever a change of region/region name occurs.

As we are introducing it to runtime, it might also be the opportunity to make it more flexible via describe-regions, and use globing as they do: `aws ec2 describe-regions --filters "Name=endpoint,Values=*us*"`
Src: http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/ & http://docs.aws.amazon.com/cli/latest/reference/ec2/describe-regions.html. This way we could easily set the lambda to run a different regions.

cloudformation/stacks/AutoSpotting/parameters.yaml
```diff
@@ -1,3 +1,8 @@
 ---
 LambdaHandlerFunction: handler.handle
 LambdaS3Bucket: cloudprowess
+LambdaMemorySize: '256'
+ExecutionFrequency: "rate(1 minute)"
+MinOnDemandNumber: '1'
+MinOnDemandPercentage: '50'
```
xlr-8: Specifying 2 default values for `MinOnDemandPercentage` & `MinOnDemandNumber` does not seem to make that much sense as Percentage will be (very likely) ignored as `MinOnDemandPercentage` is valid. This is also something we explicated in the documentation, so might be confusing for people reading the code.

Why not defaulting it to 0? as this was the default behaviour before and specify that this can be overridden in the documentation? This way, it wouldn't make a 'breaking-change' for people already using it.

autospotting.go
```diff
@@ -87,3 +90,20 @@ func readAssets() (string, []byte) {
...
 }
+
+func (c *cfgData) applyCloudWatchEventConfigs(evt json.RawMessage) {
+
+	var eventOptions autospotting.EventOptions
+
+	err := json.Unmarshal(evt, &eventOptions)
+
+	if err != nil {
+		log.Fatalln("error:", err)
```
cristim: Good catch, I'll fix this.

cloudformation/stacks/AutoSpotting/template.json
cristim: I copied it all from there, I'll fix it in both places.

cloudformation/stacks/AutoSpotting/parameters.yaml
cristim: This is just the way I execute my testing stack, it's merely an example. But you're right, I will change it to actually make some sense.

cloudformation/stacks/AutoSpotting/parameters.yaml
cristim: The code is actually doing an API call to get the full list of regions and would run everywhere, this configuration is just a filter for reducing the number of regions to a smaller subset in case the user doesn't want it to run everywhere. If this value is provided, all regions not matching those in this list would be skipped at runtime.

cristim:
@xlr-8 I ended up re-implementing this based on Lambda environment variables, which is definitely cleaner. That needed less overall code changes, basically I ended up just the replacing the standard library flag dependency with a third-party drop-in replacement that also exposes the flags as env vars, but needed more usability fixes and documentation updates.

autospotting.go
```diff
@@ -55,21 +56,28 @@ func (c *cfgData) initialize() {
...
-		"Minimum on-demand instances (percentage) running in ASG.")
+
+	flag.StringVar(&c.Regions, "regions", "",
+		"Regions where it should be activated (comma or whitespace separated list), by default it runs on all regions.")
+
+	flag.Int64Var(&c.MinOnDemandNumber, "min_on_demand_number", 0,
+		"On-demand capacity (as absolute number) ensured to be running in each of your groups. Can be overridden on a per-group basis.")
+
+	flag.Float64Var(&c.MinOnDemandPercentage, "min_on_demand_percentage", 0.0,
+		"On-demand capacity (as percentage of total capacity) ensured to be running in each of your groups. Can be overridden on a per-group basis.")
```
xlr-8: I think the term `total capacity` can be miss leading, as the percentage is based on the currently running instances.

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -26,6 +31,21 @@
...
+    },
+    "MinOnDemandPercentage": {
+      "Default": "0.0",
+      "Description": "Minimum on-demand instances (in percentage of the group capacity) that will be kept running in your ASGs. It is also a global default value that can be overridden on a per-group level using the autospotting_on_demand_percentage tag. MinOnDemandNumber takes precedence if both these parameters are passed",
+      "Type": "Number"
+    },
+    "Regions": {
+      "Default": "",
+      "Description": "Space separated list of regions where it should run, in case you may want to limit it to a smaller set of regions. If unset it will run against all available regions.",
+      "Type": "String"
```
xlr-8: Despite giving the possibility to have both space/comma separated regions, I would still stick to comma, as documented in the README, what do you think?

README.md
```diff
@@ -178,12 +178,38 @@ Notes:
 * The AutoScaling groups it runs against can be in any region, since all regions
   are processed at runtime.
 
+#### Installation options ####
+
+The CloudFormation stack takes a number of parameters, which allow you to
```
xlr-8: I think it should be `which allows` as it refers to `a number of parameters`; but could be wrong.

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -51,24 +69,20 @@
...
-          "Fn::GetAtt": [
-            "LambdaExecutionRole",
-            "Arn"
-          ]
+        "Environment": {
+          "Variables": {
+            "MIN_ON_DEMAND_NUMBER": { "Ref": "MinOnDemandNumber" },
+            "MIN_ON_DEMAND_PERCENTAGE": { "Ref": "MinOnDemandPercentage" },
+            "REGIONS": { "Ref": "Regions" }
+          }
```
xlr-8: Indeed much cleaner!

cloudformation/stacks/AutoSpotting/template.json
cristim: I tried that, but CloudFormation fails to set any value containing commas: http://docs.aws.amazon.com/lambda/latest/dg/API_Environment.html#SSS-Type-Environment-Variables

Commas are nicer when defined for manual local execution, since they don't need quotes.

autospotting.go
cristim: I don't understand what you mean, please propose a phrasing that you would find better.

README.md
cristim: I'm not a native English speaker, so I may be wrong too, but now that you mentioned this I think you are right, I'll fix this.

cristim:
I'll try to implement the region name globing, that shouldn't be so hard.

cristim:
@xlr-8 have a look at the current state of this PR.

README.md
```diff
@@ -134,6 +134,8 @@ then be terminated.
   * it tries to diversify the instance types.
   * when having enough desired capacity, it is often spreading over four
     different spot pricing zones (instance type/availability zone combinations).
+  * supports keeping a configurable number of on-demand instances in the group,
+    either an absolute number or a percentage of the instances from the group.
```
cristim: Good point, thanks!

README.md
```diff
@@ -178,12 +180,38 @@ Notes:
 * The AutoScaling groups it runs against can be in any region, since all regions
   are processed at runtime.
 
+#### Installation options ####
```
cristim: How do I do that?

autospotting.go
```diff
@@ -2,10 +2,11 @@ package main
 
 import (
 	"encoding/json"
-	"flag"
 	"log"
 	"os"
 
+	"github.com/namsral/flag"
```
cristim: Yes, a go get is unfortunately needed, I'll add it to the Makefile.

I would prefer not to use a vendoring tool for now, especially if it requires all the dependencies to be added to our git repo.

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -34,17 +34,17 @@
...
       "Default": "0",
-      "Description": "Minimum on-demand instances (number) running in ASG.",
+      "Description": "Minimum on-demand instances (absolute number) to be kept running in your ASGs. It is a global default value that can be overridden on a per-group level using the autospotting_on_demand_number tag. It takes precedence over MinOnDemandPercentage, so it doesn't make sense to pass both of them.",
       "Type": "Number"
     },
     "MinOnDemandPercentage": {
-      "Default": "0",
-      "Description": "Minimum on-demand instances (percentage) running in ASG.",
+      "Default": "0.0",
+      "Description": "Minimum on-demand instances (in percentage of the group capacity) that will be kept running in your ASGs. It is also a global default value that can be overridden on a per-group level using the autospotting_on_demand_percentage tag. MinOnDemandNumber takes precedence if both these parameters are passed",
```
cristim: Got it.

README.md
cristim: I figured it out, fixed in the current code.

autospotting.go
xlr-8: Noted

PR #64, opened 2016-12-25
Fix versioning of binaries, missing since build 108 due to a regression

cristim:


build_helper.sh
```diff
@@ -58,9 +58,6 @@ function prepare_upload_data {
   rm -rf ${LOCAL_PATH}
   mkdir -p ${LOCAL_PATH}
   mv handler.zip ${LOCAL_PATH}/lambda.zip
-}
-
-function prepare_cloudformation_code {
```
cristim: I introduced this function while refactoring the Makefile a few weeks back, but it was never called and now it doesn't make sense anymore.

xlr-8: Not sure to follow, if the function is never called, why not removing all of its content?
Also why having build_helper.sh? I feel like the code of the script, could easily go in the Makefile

cristim: The function's code is needed for the versioning, but it was just never called due to a human error, so I decided to merge its code into the function above it that is getting called. Deleting those three lines simply merges the code to the function above it.

This stuff used to be in the Makefile before, but I thought it would be cleaner to have it in a dedicated shell script.

This is how the Makefile used to look before I changed it a few weeks back: https://github.com/cristim/autospotting/blob/d51b65b01f71f152981388a44e7cab33c742d0a5/Makefile

roeyazroel:
This is not in master yet right? right now the latest binary is still build 108.

cristim:
@roeyazroel It needs to pass the code review before it can be merged.

build_helper.sh
xlr-8: OK, well then better merge it.

The Makefile can be discussed later I believe.
I do have a preference for the Makefile over scripting, as it seems more logical to use it when dealing with compiled languages, as the project/collaborator grows, you might not want to have to do full compilations all the time for a minor edit; which is something a script isn't able to do easily.
The "scripting part" (condition/loop/output management/etc) is also something that can be done via the Makefile, so I don't really feel the need to write a script for that.

cristim:
@roeyazroel it's there now.

PR #65, opened 2016-12-27
README updates

cristim:
Added an animated GIF showing the replacement workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,36 +29,55 @@ Table of Contents
 # AutoSpotting #
 
 A simple tool designed to significantly lower your Amazon AWS costs by
-automating the use of the [spot market](https://aws.amazon.com/ec2/spot). It can
-often achieve savings up to 90% off the usual on-demand prices, like shown in
-the screenshot below.
+automating the use of the [spot market](https://aws.amazon.com/ec2/spot).
 
-When enabled on your existing on-demand AutoScaling group, it starts launching
+When enabled on an existing on-demand AutoScaling group, it starts launching
 EC2 spot instances that are cheaper, at least as powerful and configured as
 closely as possible as your existing on-demand instances.
 
-It then gradually swaps them with your existing on-demand instances, which can
-then be terminated.
+It then gradually swaps them one by one with your existing on-demand instances,
+just like shown below:
+
+![Workflow](https://cdn.cloudprowess.com/images/autospotting.gif)
+
+In this case the initial instance type was quite expensive, so the algorithm
+chose a different type that had more computing capacity. At the end that group
+had 3x more CPU cores and 66% more RAM than in the initial state of the group,
+and all this with 33% cost savings and without running entirely on spot
+instances, since some users find that a bit risky.
+
+Nevertheless, AutoSpotting tends to be quite reliable even on all-spot
+configurations (has automated failover to on-demand nodes and spreads over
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,16 +29,28 @@ Table of Contents
...
+In the above example:
+- it was enabled on a group of three instances.
+- configured to keep an on-demand instance running at all times.
+- each of the two spot instances were roughly 2x cheaper, had 4x more CPU
+  cores and 2x more RAM that the on-demand instances they replaced.
+- overall we got 3x more CPU cores and 66% more RAM than we had initially, and
+  all this with 33% cost savings.
+
+In all-spot configurations, it can often achieve savings up to 90% off
+the usual on-demand prices, like shown in the screenshot below:
```
xlr-8: Just for the reference, do you know by chance, which instance type were replaced, and by what?

README.md
```diff
@@ -29,16 +29,28 @@ Table of Contents
...
+
+![Workflow](https://cdn.cloudprowess.com/images/autospotting.gif)
+
+In the above example:
+- it was enabled on a group of three instances.
+- configured to keep an on-demand instance running at all times.
+- each of the two spot instances were roughly 2x cheaper, had 4x more CPU
+  cores and 2x more RAM that the on-demand instances they replaced.
+- overall we got 3x more CPU cores and 66% more RAM than we had initially, and
+  all this with 33% cost savings.
```
xlr-8: Cool, I really like the idea!

3 small notes:
* If we keep the explanations separated, I'd put them before the gif, to avoid clicking it to maximize it, without understanding what's in there.
* It might be a bit fast even for someone who knows the project. I had trouble to follow the first time I opened it without reading the description
* If we skip the explanations separated, do you have a way to integrate it in the gif itself, by highlighting, adding description or pointing things?
e.g.: Highlighting or putting description for: Cloudformation defaults variables, Tags variables, Old instance being removed? Basically the key element of the description, that would make the gif almost self-explanatory.

cristim: Thanks, I'll create a new GIF with some descriptions embedded in it.

README.md
cristim: It can be seen in the last frame of the gif, m3.medium replaced with c3.xlarge

xlr-8: Yup, I was referring for the 'all-spot' configuration. Might be worth mentioning it that it was m3.medium replaced, by the same type in case other people wondered as well

README.md
cristim: I updated the gif, now waiting 10s between slides and with some text highlighting the important items.

README.md
cristim: Have a look at the latest text, I tried to make this clear now.

README.md
```diff
@@ -29,36 +29,55 @@ Table of Contents
...
 closely as possible as your existing on-demand instances.
 
-It then gradually swaps them with your existing on-demand instances, which can
-then be terminated.
+It then gradually swaps them one by one with your existing on-demand instances,
+just like shown below:
+
+![Workflow](https://cdn.cloudprowess.com/images/autospotting.gif)
+
+In this case the initial instance type was quite expensive, so itthe algorithm
```
xlr-8: typo: `itthe algorithm`

cristim: Good catch, thanks!

PR #68, opened 2016-12-29
Codeclimate fixes

cristim:
Trying to clean up some of the issues reported on [codeclimate](https://codeclimate.com/github/cristim/autospotting)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,7 @@ This was seen on a group of two m3.medium instances running in eu-west-1:
     down-payment or long term commitment.
 
 * **Easy to install and set up on existing environments based on AutoScaling**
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -247,7 +246,7 @@ create-or-update-tags \
 --tags ResourceId=my-auto-scaling-group,ResourceType=auto-scaling-group,Key=spot-enabled,Value=true,PropagateAtLaunch=false
 ```
 
-**Note**
+##### Note #####
```
xlr-8: This note isn't added to the TOC, which could be fine, but would be problematic if using automatic TOC generators.

cristim: Thanks, maybe it shouldn't be a section after all.

PR #71, opened 2017-01-05
makefile: re-implementing logic into the Makefile

xlr-8:
To avoid having scripts and Makefile, the logic has been re-implemented
at the Makefile level, thus allowing logic between rules, compile only
the necessary parts, etc.

Parts for go vet / fmt and help have also been added.

Note: Comments/Feedback appreciated.


Makefile
```diff
@@ -1,28 +1,91 @@
...
+
+html-cover:                                                  ## Display coverage in HTML
+	@go tool cover -html=$(COVER_PROFILE)
+
+help:                                                        ## Show this help
+	@echo -e "Rules:\n"
+	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
+
+.PHONY: all clean check_deps bindata build_lambda_binary prepare_upload_data build_local upload vet-check fmt-check
+.PHONY: test full-test html-cover help
```
cristim: No newline ant the end of file

Makefile
```diff
@@ -1,28 +1,91 @@
...
-build_local: prepare_bindata
-	./build_helper.sh build_local
+all: build_local                                             ## Build the code
 
-upload: prepare_upload_data
-	./build_helper.sh upload
+clean:                                                       ## Remove installed packages/temporary files
+	go clean ./...
+	rm -rf $(BINDATA_DIR) $(BINDATA_FILE)
+	docker rmi -f $(DOCKER_IMG) 2> /dev/null || true
```
cristim: I wouldn't remove the image here, but that's not a big deal.

Makefile
```diff
@@ -1,28 +1,91 @@
...
+	go build $(GOFLAGS) -o $(BINARY)
+
+upload: prepare_upload_data                                  ## Upload binary
+	aws s3 sync build/s3/ s3://$(BUCKET_NAME)/
+
+vet-check:                                                   ## Verify vet compliance
+ifeq ($(shell go tool vet -all -shadow=true $(GO_FILES) 2>&1 | wc -l), 0)
+	@echo -e "ok\tall files passed go vet"
+else
+	@echo -e "ko\tsome files did not pass go vet"
```
cristim: what is ko?

Makefile
```diff
@@ -1,28 +1,91 @@
...
+	@go tool vet -all -shadow=true $(GO_FILES) 2>&1
+endif
+
+fmt-check:                                                   ## Verify fmt compliance
+ifneq ($(shell gofmt -l $(GO_FILES) | wc -l), 0)
+	@echo -e "ko\tsome files did not pass go fmt: $(shell gofmt -l $(GO_FILES))"; exit 1
```
cristim: again ko?

generated_bindata.go
```diff
@@ -89,7 +89,7 @@ func dataBuild() (*asset, error) {
...
...8\xc4\x0b\x44\xa7\xeb\x39\xac\x15\x31\x57\x14\x3c\x09\xaa\x7f\x70\x16\x48\xab\xe8\x21\x2d\x4b\x2b\x62\x4b\x23\x3c\xbb\xa7\xc9\x8a\x72\xc1\xdf\x40\x0e\x45\x06\x5b\x0c\xc8\x86\x9b\xfd\x0a\xdc\x66\x31\x20\xaa\x72\x1a\x36\xa4\xda\xb0\xd1\xc2\xd0\xca\xc6\x1c\xa2\x1e\xa1\xc9\x6f\x4b\x42\xfe\x83\x97\x60\x68\xd9\x95\xc2\x6d\x39\xc4\xa6\xbf\xa7\x77\x64\xb6\x90\x3b\x8d\xfa\x7b\x7a\x86\x96\x82\xdc\x09\x26\x64\x97\x60\x68\x99\xcb\x89\x68\xcb\x21\x08\x05\xa3\x19\x5a\x7a\xf3\x4b\x5e\xc8\x5c\x0e\xed\x1d\xce\xd0\x92\xab\x72\x7a\xce\xed\x6c\xe8\x60\x1d\xd5\x68\x82\x3e\xc4\xa6\xe1\x8a\xb0\x05\xba\x0f\x51\xdf\x16\xb8\x42\x47\xf5\x21\x26\x53\xa3\xc8\xb7\xc8\x4d\x1f\x62\xdc\x53\xbd\x70\x0e\xb1\x2e\x1d\x9a\x37\xcd\x0c\x36\x89\x98\xc6\x43\x44\x01\xc3\xca\x32\xfb\x5b\xfa\xe5\xb4\x1a\xce\x52\x68\xb8\x96\x81\x39\xc4\xbc\x74\x88\x58\x6d\x5c\x1f\x62\x18\x94\x06\xa4\xe2\x4f\xe5\x10\xbd\xf1\x08\x64\x02\x31\x7d\x88\x1a\xce\x92\xbb\xe8\xc6\xae\xe7\xf2\x77\x37\x62\x8b\x81\x72\x9b\x6f\x9e\x99\xc5\xc0\xdf\x54\x1e\x03\xa1\x3e\x92\xc9\xc0\xfa\xf5\x1f\x1f\xdf\x64\xe0\xef\xdf\x7f\xfe\xcb\x9b\x57\x57\xea\x35\x50\xa0\x3b\x32\x17\x87\x0a\x4f\xa7\xea\x3e\x66\x41\x49\x0b\xbf\x12\xb2\x7f\xb1\xcc\x3b\xb1\x34\x67\x61\xab\x4e\xb
```
cristim: What changed here? Let's keep this file out of this PR, or maybe submit it independently in another PR.

Makefile
```diff
@@ -1,28 +1,91 @@
...
+	docker run --rm -v $(GOPATH):/go -v $(PWD):/tmp $(DOCKER_IMG)
+
+prepare_upload_data: build_lambda_binary                     ## Create archive to be uploaded
+	@rm -rf $(LOCAL_PATH)
+	@mkdir -p $(LOCAL_PATH)
+	@mv handler.zip $(LOCAL_PATH)/lambda.zip
+	@cp -f cloudformation/stacks/AutoSpotting/template.json $(LOCAL_PATH)/template.json
+	@cp -f cloudformation/stacks/AutoSpotting/template.json $(LOCAL_PATH)/template_build_$(BUILD).json
+	@cp -f $(LOCAL_PATH)/lambda.zip $(LOCAL_PATH)/lambda_build_$(BUILD).zip
+	@cp -f $(LOCAL_PATH)/lambda.zip $(LOCAL_PATH)/lambda_build_$(SHA).zip
```
cristim: I'm thinking to either drop the SHA entirely (also needs doc cleanup) or also do the SHA also for the CloudFormation template. What do you think about this?

Makefile
```diff
@@ -1,28 +1,91 @@
...
+BUILD := $(or $(TRAVIS_BUILD_NUMBER), $(TRAVIS_BUILD_NUMBER), $(SHA))
 
-# the following targets are only used for local development
+# upstream data
+EC2_INSTANCES_INFO_COMMIT_SHA := e655e36660bc617713c4ef9a1409cc65a209cb27
+INSTANCES_URL := "https://raw.githubusercontent.com/powdahound/ec2instances.info/$(EC2_INSTANCES_INFO_COMMIT_SHA)/www/instances.json"
 
-build_local: prepare_bindata
-	./build_helper.sh build_local
+all: build_local                                             ## Build the code
```
cristim: Why double the # for comments?

Makefile
```diff
@@ -1,28 +1,91 @@
...
+	done
+	@echo "all dependencies satisifed: $(DEPS)"
 
-cover:
-	./build_helper.sh calculate_coverage
\ No newline at end of file
+bindata: check_deps                                          ## Convert instance data into go file
+	@type go-bindata || go get -u github.com/jteeuwen/go-bindata/...
+	@mkdir -p $(BINDATA_DIR)
+	wget --quiet -nv -c $(INSTANCES_URL) -O $(BINDATA_DIR)/instances.json
```
cristim: wget -c is problematic and may corrupt the JSON, let's drop it.

I know it would be faster but I'd rather wait a second more instead of having broken builds... unless we can find a way to check the ETAG before getting the data?

Perhaps this may work out: http://blog.yjl.im/2012/03/downloading-only-when-modified-using.html

Makefile
xlr-8: See make help:
```bash
> make help
Rules:

all: build_local                                              Build the code
clean:                                                        Remove installed packages/temporary files
check_deps:                                                   Verify the system has all dependencies installed
bindata: check_deps                                           Convert instance data into go file
build_lambda_binary: bindata                                  Build lambda binary
prepare_upload_data: build_lambda_binary                      Create archive to be uploaded
build_local: bindata                                          Build binary - local dev
upload: prepare_upload_data                                   Upload binary
vet-check:                                                    Verify vet compliance
fmt-check:                                                    Verify fmt compliance
test:                                                         Test go code and coverage
full-test: test fmt-check vet-check                           Pass test / fmt / vet
html-cover:                                                   Display coverage in HTML
help:                                                         Show this help
```

Makefile
xlr-8: Simply the error message:
```bash
ok       all files passed go vet # In success
ko       some files did not pass go vet # In failure
```

Makefile
xlr-8: See above

generated_bindata.go
xlr-8: Noted

Makefile
```diff
@@ -1,28 +1,91 @@
...
-	./build_helper.sh prepare_bindata
+GO_FILES := $(shell find ./ -name '*.go')
 
-build_lambda_binary: prepare_bindata
-	./build_helper.sh build_lambda_binary
+BINDATA_DIR := data
+BINDATA_FILE := generated_bindata.go
 
-prepare_upload_data: build_lambda_binary
-	./build_helper.sh prepare_upload_data
```
xlr-8: These rules are missing and need to be re-implemented

Makefile
xlr-8: Will be dropped.

Makefile
xlr-8: Hum, which were the reasons for each of those? This is something I'm not quite familiar yet.

First thoughts would be that I'd drop the BUILD number to keep the SHA only, in order to have something unique for every commit, rather than a BUILD number which can be problematic - e.g. recently the increase was broken.

Not sure when this is 'relevant' to know which version is running, and of course it's not as "friendly" as a build number, but I'd rather have something unique and functioning despite being not too friendly, than something that could break.

Makefile
cristim: Sweet!

Makefile
cristim: Understood, I'm fine either way, your call.

Makefile
cristim: This is for allowing people to easily reason about the version that they use and switch to another one, I prefer to have the numbers for ease of use, I also use them in the changelog entries. 

The numeric build numbers are all generated by Travis, and are always increasing, including for pull request verification builds, but those will not be pushed to S3 so there will always be some missing numbers and the bucket has to be checked for availability of a certain build, as documented.

The SHA is for custom builds that are not pushed through Travis CI (mainly for testing) and/or installed through another S3 bucket.

The broken build number issue is unrelated, it would have also been broken for SHA because this entire section was not executed at all.

Makefile
cristim: I would use an English word or irs abreviation, such error (err), warning (warn)

xlr-8: That was to give such outputs:
```bash
# All OK
>make full-test
ok  	github.com/cristim/autospotting/core	0.008s	coverage: 21.3% of statements
ok	all files passed go fmt
ok	all files passed go vet
# Fmt fail
> make full-test
ok  	github.com/cristim/autospotting/core	0.009s	coverage: 21.3% of statements
ko	some files did not pass go fmt: ./autospotting.go
make: *** [Makefile:72: fmt-check] Error 1
# Test fail
> make full-test
--- FAIL: TestGetTagValue (0.00s)
    --- FAIL: TestGetTagValue/Tag_can_be_found_in_ASG_(many_tags) (0.00s)
    	autoscaling_test.go:67: Value received for spot-enabled: true expected false
FAIL
coverage: 21.3% of statements
FAIL	github.com/cristim/autospotting/core	0.009s
make: *** [Makefile:78: test] Error 1
```

cristim: I find it hard to see, since inversions of the same characters are not very visible.

How about some ANSI color codes for green (ok) and red (failure)?

cristim: Or just make the entire text of the errors red, without any such prefix?

Makefile
xlr-8: Oh okay, sorry for the misunderstanding, then I'm fine with either way. As you said build number might be easier/more readable, we can drop the SHA and see later if it doesn't cause problem. But might be better to do that in another PR I believe.

Makefile
xlr-8: I think colour would lead to a more difficult reading in the makefile itself, but better in the terminal.

I honestly don't really see the point of having error/warning message instead of ok/ko, as this is mostly for the CI, so the feedback would be done on the commit, and the last message would be the one failing as the makefile exits as the first error.
For local testing, I have the same observation, the makefile message `Error 1` is quite obvious.

But I can change it, don't really mind. I'll put error if that's fine.

build_helper.sh
```diff
@@ -1,106 +0,0 @@
...
-  for DEP in $DEPS
-  do
-    check_dep $DEP && echo "Found ${DEP}"
-  done
-  echo "All dependencies were successfully found, we're good to go!"
-}
-
-function prepare_bindata {
-  check_all_dependencies
-  go get ./...
```
xlr-8: Was this line necessary?

Makefile
xlr-8: Done

build_helper.sh
cristim: I think that's needed in case the software is cloned manually from github instead of using `go get`

Makefile
```diff
@@ -1,28 +1,90 @@
...
+	@for DEP in $(shell echo "$(DEPS)"); do \
+		command -v "$$DEP" &>/dev/null \
+		|| (echo "Error: dependency '$$DEP' is absent" ; exit 1); \
+	done
+	@echo "all dependencies satisifed: $(DEPS)"
 
-cover:
-	./build_helper.sh calculate_coverage
\ No newline at end of file
+prepare_bindata: check_dep                                  ## Convert instance data into go file
```
cristim: The "Convert instance data into go file" comment is misaligned with the rest, needs an extra whitespace before.

I would expect make to be smart enough to align them all.

Makefile
cristim: Let's keep both for now

cristim:
@xlr-8 looks good to me.

Do you want to iterate more or can I merge this now?

xlr-8:
If you double checked the rules, I think you can merge it. Something that could be good, would be to integrate the full-test into the CI to have also fmt and vet being checked.

cristim:
```
$ make full-test 
ok  	github.com/cristim/autospotting/core	0.006s	coverage: 21.3% of statements
-e ok	all files passed go fmt
-e ok	all files passed go vet
```
Any idea why does it print the '-e' ?

cristim:
Deleting the -e did the trick
```
make full-test
ok  	github.com/cristim/autospotting/core	0.006s	coverage: 21.3% of statements
ok	all files passed go fmt
ok	all files passed go vet
```

xlr-8:
The `-e` is to have the escape enabled, otherwise I have the `\t` being displayed:
```bash
>make full-test
ok  	github.com/cristim/autospotting/core	0.005s	coverage: 21.3% of statements
ok\tall files passed go fmt
ok\tall files passed go vet
```
Do you have an alias on the echo command?

cristim:
That's strange, I don't have such an alias.

cristim:
I replaced `echo` with `printf` wherever it needed to handle escape sequences, since echo is not really portable across shell implementations.

@xlr-8 please confirm if it works now.

xlr-8:
I just understood why the echo might have been a problem. What is the output of: `ls -l /bin/sh` ?
On my side it is `/bin/bash`, but maybe on yours it's something different which is used by default by the Makefile, thus resulting in a different behavior on each side.

cristim:
I am using a ZSH, likely to be slightly incompatible. 

I still think the portable way is to use printf.

Issue #62, opened 2016-12-25
Improve the instance replacement logic


cristim:
The code needs to be cleaned up at least as per some of the code review comments posted on #46.

Any function that is changed needs to have its unit tests updated or created if missing.

PR #79, opened 2017-01-31
Improve replacement logic

xlr-8:
This PR updates the replacement logic which was previously reported.
It also includes a set of tests for almost all methods of instance.go (~95% code coverage).

The infinite loop for tagging has been updated in order to take a timeout in parameter (default value 5min with wait of 5sec interval). It's a long time, it can be customized, it's not ideal, but it's still better than the infinite loop. An issue could be open in order to improve the retry logic, as it was previously mentioned as well.

The test for `getCheapestCompatibleSpotInstanceType` is not particularly crazy, but there are so many cases that would have to be treated that 2 simple have been put, while all 'logical methods' from this methods are being tested properly. It might not make sense to add more tests to it.

Would solve: https://github.com/cristim/autospotting/issues/62 & https://github.com/cristim/autospotting/issues/42


core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			i := &instance{Instance: &ec2.Instance{}}
+			i.InstanceLifecycle = tt.lifeCycle
+			retValue := i.isSpot()
+			if retValue != tt.expected {
+				t.Errorf("Value received for '%s': %s expected %s", tt.lifeCycle, retValue, tt.expected)
```
cristim: tt.lifeCycle is a *string, can't be printed with %s

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
+package autospotting
```
cristim: I get a complaint about not passing go fmt

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+				Placement: &ec2.Placement{
+					AvailabilityZone: tt.availabilityZone,
+				}},
+				price: tt.instancePrice,
+			}
+			candidate := instanceTypeInformation{pricing: prices{}}
+			candidate.pricing = tt.spotPrices
+			retValue := i.isPriceCompatible(candidate, tt.bestPrice)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: tt.expected is a bool and should be printed with %t

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			i := &instance{typeInfo: instanceTypeInformation{
+				vCPU:   tt.instanceCPU,
+				memory: tt.instanceMemory,
+			},
+			}
+			retValue := i.isClassCompatible(tt.spotInfo)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: again we should use %t when printing bool variables

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+			expected:        true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			i := &instance{typeInfo: tt.instanceInfo}
+			retValue := i.isStorageCompatible(tt.spotInfo, tt.attachedVolumes)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: and once more %t

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			i := &instance{Instance: &ec2.Instance{
+				VirtualizationType: tt.instanceVirtualizationType,
+			}}
+			retValue := i.isVirtualizationCompatible(tt.spotVirtualizationTypes)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: once again

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+				Instance: &ec2.Instance{
+					Placement: &ec2.Placement{
+						AvailabilityZone: tt.availabilityZone,
+					},
+				},
+				asg: a,
+			}
+			retValue := i.isSpotQuantityCompatible(tt.spotInfo)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: again

core/instance_test.go
```diff
@@ -0,0 +1,485 @@
...
+			x:        0,
+			y:        10,
+			expected: 0,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			retValue := min(tt.x, tt.y)
+			if retValue != tt.expected {
+				t.Errorf("Value received: %s expected %s", retValue, tt.expected)
```
cristim: int values need to be printed with %d

xlr-8:
Thanks, I fixed the various comments, still need to implement the last method tests.

xlr-8:
Yes, we could.
Although here it was my fault, as I also have a local hook that I bypassed during a local rebase. Should have done a `full-test` instead of `test` also :)

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
+	}
+}
+
+func TestAdd(t *testing.T) {
+	tests := []struct {
+		name     string
+		catalog  map[string]*instance
+		expected map[string]*instance
+	}{
+		{name: "map contain a nil pointer",
```
cristim: contains

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
+		},
+		{name: "map has 1 instance",
+			catalog: map[string]*instance{
+				"inst1": &instance{Instance: &ec2.Instance{InstanceId: aws.String("1")}},
+			},
+			expected: map[string]*instance{
+				"1": &instance{Instance: &ec2.Instance{InstanceId: aws.String("1")}},
+			},
+		},
+		{name: "map has several instance",
```
cristim: instances

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
+			catalog:  map[string]*instance{},
+			expected: 0,
+		},
+		{name: "map has 1 instance",
+			catalog: map[string]*instance{
+				"id-1": {},
+			},
+			expected: 1,
+		},
+		{name: "map has several instance",
```
cristim: instances

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
```
cristim: instances

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
+func TestIsPriceCompatible(t *testing.T) {
+	tests := []struct {
+		name             string
+		spotPrices       prices
+		availabilityZone *string
+		instancePrice    float64
+		bestPrice        float64
+		expected         bool
+	}{
+		{name: "Not spot price for such availability zone",
```
cristim: "No"  instead of "Not"

core/instance_test.go
```diff
@@ -0,0 +1,1027 @@
...
+					"eu-west-1":    1.0,
+					"eu-west-2":    2.0,
+				},
+			},
+			availabilityZone: aws.String("eu-west-1"),
+			instancePrice:    5.0,
+			bestPrice:        1.4,
+			expected:         true,
+		},
+		{name: "Spot price is equal 0.0",
```
cristim: equal could be skipped

xlr-8:
Should be all fixed

Issue #41, opened 2016-11-21
Write tests for autoscaling.go


cristim:

xlr-8:
The coverage is now at 90% on this file, so closing this one as well.
ref: https://coveralls.io/github/cristim/autospotting?branch=master

Issue #7, opened 2016-08-19
Increase the automated test coverage to an acceptable value


cristim:
I'd like to see it achieve somewhere around 80%

Summary of files:
* [x] autoscaling (ref https://github.com/cristim/autospotting/issues/41)
* [x] connections (no ref)
* [x] instance (ref https://github.com/cristim/autospotting/issues/42)
* [x] launch_configuration (ref https://github.com/cristim/autospotting/issues/43)
* [ ] main (no ref)
* [ ] region (ref https://github.com/cristim/autospotting/issues/44)
* [x] spot_instance_request (ref https://github.com/cristim/autospotting/issues/45)
* [x] spot_price (no ref)

xlr-8:
I guess we can close it now: https://coveralls.io/github/cristim/autospotting?branch=master

Except `region.go` & `main.go` the coverage of the other one has been reached, and overall coverage as well.

cristim:
Thanks, I'll close this.

But I'd like to also increase the coverage on region.go and main.go at some point.

xlr-8:
Agreed, which is why I created a new one, and update this description.

PR #89, opened 2017-02-28
Test autoscaling

xlr-8:
# Issue Type #

- Improvement test Pull Request

Should help with:
 * https://github.com/cristim/autospotting/issues/41
 * https://github.com/cristim/autospotting/issues/7

## Summary ##

Most of the methods have been covered by this PR - roughly 65-70% is covered now.
Some methods have been modified to be easier to test (return of values, etc), some have been merged when possible, and some have been removed when not used, please refer to commit/commit's message for more details.

Other bigger methods however are not yet covered by this PR, I believe they might need to be modified/split/simplified to have a better testing/understanding/reading of them. Those are:
 * [process](https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L171)
 * [havingReadyToAttachSpotInstance](https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L398)
 * [launchCheapestSpotInstance](https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L511)

I believe this should be a good start to cover this file.

Note: Still considered WIP as I'd like to review it all myself, and do further tests of the code before merging, external reviews are however welcome as well.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"errors"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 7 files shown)


core/autoscaling.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"errors"
```
xlr-8: Yes full-test was run on the PR

xlr-8: I just realised, it might be because you used goimports while I was using go fmt in my hook. Which is why the formatting were fighting

PR #92, opened 2017-03-02
Run fmt and vet from TravisCI, build cleanups

cristim:
# Issue Type #

- Feature Pull Request

## Summary ##

This ensures the code fails build verification in case it fails to pass go vet and go fmt. 
It also creates a few additional make targets called from TravisCI so that the Travis config file can be simplified and all the installation of dependencies and build logic is in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,6 @@ LOCAL_PATH := build/s3/dv
 
 DOCKER_IMG := eawsy/aws-lambda-go
 
-GO_FILES := $(shell find ./ -name '*.go')
-
 BINDATA_DIR := data
 BINDATA_FILE := generated_bindata.go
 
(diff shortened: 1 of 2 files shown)


cristim:
Thanks, I'll update it ASAP

Makefile
```diff
@@ -54,38 +59,46 @@ prepare_upload_data: build_lambda_binary                     ## Create archive t
...
+
+travisci: prepare_bindata travisci-checks prepare_upload_data travisci-cover ## Executed by TravisCI
+
+help:                                                                        ## Show this help
 	@printf "Rules:\n"
 	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
 
-.PHONY: all clean check_dep bindata build_lambda_binary prepare_upload_data build_local upload vet-check fmt-check
-.PHONY: test full-test html-cover help
+.PHONY: all clean check_dep bindata build_lambda_binary prepare_upload_data build_local upload vet-check fmt-check check_deps prepare_bindata test full-test html-cover help travisci-checks travisci-cover travisci
```
xlr-8: Notes:
* bindata rule doesn't seem to exist anymore - could be removed from build_local / build_lambda_binary
* the check_dep rule can be removed - as renamed check_deps

xlr-8: The .PHONY could also be duplicated on every rule to make it easier/more obvious when modifying, such as:

```bash
.PHONY: rule1
rule1: dep1
   xxxx

.PHONY: rule2
rule2: dep2
   xxxx
```

PR #93, opened 2017-03-05
Switch build system to easwy/aws-lambda-go-shim

cristim:
# Issue Type #

- Feature Pull Request

## Summary ##

This build switches to eawsy/aws-lambda-go-core/shim, which is now ready for production, since the old easwy/aws-lambda-go is now deprecated.

As a requirement due to a golang limitation, it needed less symbols in the main package so I had to move generated_bindata into a dedicated ec2instancesinfo library. I was planning to do this for a long time in order to potentially have more people use this data, but now I had no other choice. Once at it, I also updated the instance information to the latest data available upstream.

The code and build system was slightly refactored in order to accommodate these changes and a few changes were done here and there, the most notable being the passing of the version string as a linker flag instead of an asset.


cristim:
The missing version number in the lambda build is a manifestation of https://github.com/golang/go/issues/19418, and it will be fixed ASAP.

cristim:
I'd like to get this merged, can someone review it?

PR #96, opened 2017-03-25
fix typo in terraform example

phils:
"rate(1 minutes)" is invalid and isn't accepted by the API.  Should be "rate(1 minute)"

### DIFF
--- a/START.md
+++ b/START.md
@@ -119,7 +119,7 @@ module "autospotting" {
   lambda_runtime = "200"
   lambda_memory_size = "2048"
   lambda_timeout = "600"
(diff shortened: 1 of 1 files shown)


phils:
Yeah, 5 minutes would be fine I think, and fits with the defaults.  I wasn't sure whether people tend tup increase the frequency when they start using the script for important things.

cristim:
I've seen people use 1 minute when testing, which makes sense if you want to quickly see how it works, but for actual usage I would recommend to use the 5 minutes default frequency.

It is still fast enough, but can be safer in case of misconfigurations, like if the group has the grace period accidentally set to zero.

PR #97, opened 2017-03-25
Handler case fix

phils:
# Issue Type #

- Bugfix Pull Request
- Documentation Pull Request

## Summary ##

Wasn't able to get this module working using the included terraform code due to error:
```
Handler 'handler' missing on module 'handler': plugin: symbol handler not found in plugin lambda
```

Digging into the code and comparing to the Cloudformation definition, I could see it is actually defined as "Handle" rather than "handle".  Fixing this allowed it to work.  Also saw reference to "handle" in the documentation so have also tweaked that.

### DIFF
--- a/START.md
+++ b/START.md
@@ -286,7 +286,7 @@ Travis CI [builds page](https://travis-ci.org/cristim/autospotting/builds)
   binary. Make sure you use the following stack parameter on any newer builds:
 
 ```
-LambdaHandlerFunction: handler.handle
+LambdaHandlerFunction: handler.Handle
 ```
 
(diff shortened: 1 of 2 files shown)


xlr-8:
Good to know, for the reference it was introduced since the refactoring https://github.com/cristim/autospotting/pull/93. Something to keep in mind if this part is *unlikely* modified again.

cristim:
I hope we won't need to change this anytime soon, I only expect it to happen again once AWS offers native golang support in case it will require a different way to configure the handler. Whenever that happens we should keep this in mind.

PR #99, opened 2017-03-29
Improve documentation of custom builds

cristim:
# Issue Type #

- Documentation Pull Request

## Summary ##

A user noticed that the custom builds documentation was inaccurate and suggested some fixes.

Once at it, I also noticed some more and I did a few more improvements.

### DIFF
--- a/CUSTOM_BUILDS.md
+++ b/CUSTOM_BUILDS.md
@@ -16,9 +16,12 @@ your CloudFormation or Terraform stack to use those new binaries.
 
 1. Verify that they were properly installed.
 
-   `go version`
+   `go version`, should be at least 1.7
+
    `git version`
+
    `docker version`
+
    `aws --version`
 
 ## Compiling the binaries locally ##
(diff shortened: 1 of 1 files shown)


ahaverbuch:
Two thoughts we saw when looking into custom builds:

1) Instructions for building from a fork (though that's a general go build scenario)
2) Are there specific docker or aws cli versions wanted/tested?

cristim:
1. Building from a fork shouldn't make any difference, that's why I didn't document it, but I may take things for granted that other people may expect differently. Do you have any concrete suggestions?

2. Not that I know of, I just run recent enough versions. But I thought the same about golang, which turned out not to be the case since for the tests we now require at least 1.7... 
I would leave them unspecified unless someone else runs into any issues.

ahaverbuch:
These were the hints we had to follow - but you pretty much need to do this on any go fork due to import paths: http://code.openark.org/blog/development/forking-golang-repositories-on-github-and-managing-the-import-path

cristim:
Wow, that's a great catch!

As I said, I took this for granted since I run the original and not a fork.

I'll point to that page from this document.

Issue #48, opened 2016-11-23
Properly handle SecurityGroups in DefaultVPC environments


cristim:
When using a stack created for EC2 classic, the groups are created by name.

In VPC(inclusing DefaultVPC) they need to be given by ID.

The code may need to query the groups by name and return their ID, and pass them by ID on VPC environments.

cristim:
I found a way to reproduce this problem:

It is not happening on resources created manually in the DefaultVPC, which was how I tried to reproduce it first, but it happens when launching a CloudFormation stack created for EC2 Classic.

xlr-8:
Could you describe the problem some more, it is not very clear to me what's the problem is, or what it affects? Maybe we could set standards on how to open issues: version, description, expected, etc

cristim:
Good point.

The problem happens on Default VPC environments when launching CloudFormation stacks created for EC2 Classic. These stacks would reference the security groups by name instead of their ID.

For some reason the application fails to launch any spot instances because of the way the security groups are being passed when creating the spot request, because the list of groups is ambiguous and the groups can be given by either name or ID, which both come from the same field of the launch configuration, and we have no way to distinguish them yet.

As a solution I was thinking to query the groups, search them by both ID and name and always use their IDs. This requires an additional IAM permission to query the groups, and I am not sure it would work on EC2 classic.

Another way may be to match them against the 'sg-*' glob and assume they are actually IDs in case of a match.

xlr-8:
Okay, thank you for the precisions, it makes more sense. Maybe the labels of this issue could be updated to add 'cloudformation' (could also be a good idea to 'terraform' or 'installers') and remove the one requiring information?

cristim:
It's not a problem with the AutoSpotting installers, but with the way the AutoScaling group was initially created.

This impacts AutoScaling groups created in DefaultVPC by stacks that were written initially for EC2 Classic, where security groups were given by name instead of by ID.

Essentially, on VPC environments AutoSpotting assumes that the security groups are referenced by ID not by name, which is not the case for DefaultVPC.

ahaverbuch:
My team found, to opposite the above, that in accounts w/ just EC2 Classic (and not DefaultVPC) the id's come back as Ids - but the code is attempting to submit the ids as SecurityGroups, not SecurityGroupIds, and thus, errors out with an error saying SecurityGroup not found. Do we have a full list of what the scenarios are between VPC, DefaultVPC, Classic?

[...]

PR #100, opened 2017-03-29
Handle #48 in short-term by using appropriate securitygroup form

ahaverbuch:
# Issue Type #

- Bugfix Pull Request

## Summary ##

Per discussion in #48, we need SecurityGroupId support for non-VPC. This flips between SecurityGroups and SecurityGroupId based off of "sg-", which while not perfect, is a good midterm solution

Confirmed it works in our scenario.

### DIFF
--- a/core/launch_configuration.go
+++ b/core/launch_configuration.go
@@ -92,8 +92,22 @@ func (lc *launchConfiguration) convertLaunchConfigurationToSpotSpecification(
 			},
 		}
 	} else {
-		// Instances are running in EC2 Classic.
-		spotLS.SecurityGroups = lc.SecurityGroups
+		// Instances are running in EC2 Classic, but maybe by name or ID
+		// depending on your scenario, so testing if start with sg-
+		// note: this doesn't yet cover scenario of mixed mode
+		ids := true
+
+		for i := range lc.SecurityGroups {
+			if !strings.HasPrefix(*(lc.SecurityGroups[i]), "sg-") {
+				ids = false
+			}
+		}
+
+		if ids {
+			spotLS.SecurityGroupIds = lc.SecurityGroups
+		} else {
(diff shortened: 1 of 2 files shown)


cristim:
I think whoever names his SGs starting with `sg-` is asking for trouble and doing it this way should be good enough for the vast majority of the users. It's also simple to implement, which is a huge plus.

We should just document this as a known issue.

xlr-8:
@cristim @ahaverbuch - sorry in advance for "being that guy" 
Just adding my 2 cents, I'm not a big fan of the 'temporary-fix' or 'mid-term solution' - if the problem isn't really tricky. From what I understood it wouldn't be that complicated to implement a management of both, no? If so, why hasn't this been done, nor issue created/documentation updated to specify the limitation of "sg-" security groups?

cristim:
@xlr-8 this was a quite nasty issue that rendered it entirely useless for a considerable number of users having that certain configuration.

In such cases I prefer to have a quick fix that works for 99+% of the users, if possible, and implement the proper fix afterwards without any pressure to deliver.

I'm not sure the proper fix is as trivial in this case as it may seem, unfortunately I didn't have to take care of it yet.

Patches are welcome, as always :-)

xlr-8:
Yes, I can understand.

Also I would mitigate the number of users impacted by it, except if you have statistics (?).
As for a matter of facts, you opened the issue in November, and nobody else "up" it, except me by curiosity, so until recently, people might have been affected without saying anything, but that didn't seem as such a "big problem" to me.

As usual, I'd be happy to help but have been pretty busy lately, hope it will go better soon enough.

cristim:
I've seen it on one of my accounts classic but didn't have time to handle it, @ahaverbuch saw it as well and I think there was someone else who mentioned it a longer while back.

Considering the currently small userbase, I think it's enough to expedite this fix to avoid first use problems with any other new users.

PR #108, opened 2017-06-19
Commenting the Manual Log Group Creation for Lambda

machbio:
# Issue Type #

- Bugfix Pull Request

## Summary ##

Terraform template for Auto-Spotting manually creates the Log group for the Lambda function. The issue with destroy and apply cycle of the terraform template - the log group is not deleted - results in error; also, the lambda functions automatically create the log group for any lambda function deployed. 

The below changes comments out the Terraform template for log group creation, as the Lambda function automatically creates the log group - if the role associated with the Lambda function has permissions to create the log group.

### DIFF
--- a/terraform/autospotting/lambda.tf
+++ b/terraform/autospotting/lambda.tf
@@ -47,9 +47,3 @@ resource "aws_cloudwatch_event_rule" "cloudwatch_frequency" {
   name = "autospotting_frequency"
   schedule_expression = "${var.lambda_run_frequency}"
 }
-
-# Commenting out - as reapplying creates an issue when manually creating the log group for Lambda
-# resource "aws_cloudwatch_log_group" "log_group_autospotting" {
-#   name = "/aws/lambda/${aws_lambda_function.autospotting.function_name}"
-#   retention_in_days = 7
-# }
--- a/terraform/autospotting/lambda.tf
+++ b/terraform/autospotting/lambda.tf
@@ -48,7 +48,8 @@ resource "aws_cloudwatch_event_rule" "cloudwatch_frequency" {
   schedule_expression = "${var.lambda_run_frequency}"
 }
 
-resource "aws_cloudwatch_log_group" "log_group_autospotting" {
-  name = "/aws/lambda/${aws_lambda_function.autospotting.function_name}"
-  retention_in_days = 7
-}
+# Commenting out - as reapplying creates an issue when manually creating the log group for Lambda
+# resource "aws_cloudwatch_log_group" "log_group_autospotting" {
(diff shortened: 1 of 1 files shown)


cristim:
I understand the problem, but that log group is defined there only in order to be able to limit the retention period to 7 days, otherwise the logs are kept forever.

I propose to create a custom resource based on the aws cli which just creates it and configures the retention period on it, and then we can remove this resource.

xlr-8:
@machbio, could you fill up an issue with logs/commands/vesions of what you describe please?

Feel-free to reference the PR as fixing the issue, it would simply be easier to discuss it rather than over a PR. On the things you mention there is: `Commenting out - as reapplying creates an issue when manually creating the log group for Lambda`, but if you create it manually, you're basically missing the point of using terraform - except if there is another reason? Why not adding it to your terraform files rather than creating it manually?

On a more general note about the PR, I feel like it would not make sense to leave code commented, whether the code is useful and can be kept, whether it needs to be fixed and un-commented.

machbio:
@cristim the whole purpose of terraform is not utilized if this pattern is followed.

@xlr-8 
Sorry - I meant, when creating Log Group through terraform script creates an issue (not manually create) .. I can also delete the code commented out

cristim:
@machbio do you have any other suggestion? 

Perhaps have a custom resource called before the deletion of the log group which would purge the log data?

xlr-8:
@machbio Could you have provide a handful of commands that you're doing in order to reproduce that error? I'd like to test it on my side as well to validate the issue.

Feel free to also drop the versions of both go, terraform, and autospotting that you're using.

cristim:
@xlr-8 I think just removing the terraform stack should be enough, as long as the function executed once and created some logs the deletion of the log group would fail if not empty

machbio:
@xlr-8 
1. Create Autospotting Infrastructure
`terraform apply`
2. Destroy Autospotting Infrastructure 
`terraform destroy -force`
3. Recreate Autospotting Infrastructure 
`terraform apply`
ERROR: Log Group already exists

@cristim that would be my suggestion - there is no way to currently create log group - https://www.terraform.io/docs/providers/aws/r/cloudwatch_log_group.html

`name - (Optional, Forces new resource) The name of the log group. If omitted, Terraform will assign a random, unique name.`

xlr-8:
I tried to reproduce your error in an empty environment, but I couldn't. Then Terraform module works as expected on my side. I created the issue - just in case - please reply there @machbio . Thanks!

cristim:
Closing since it couldn't be reproduced, @machbio please reopen if you can get someone else able to reproduce this issue.

Issue #111, opened 2017-06-20
Implement retry logic for tagging of spot instance requests


cristim:
This hopefully fixes #20

xlr-8:
As AWS announced in the past couple months, tag can now directly be transmitted to the instances, so that we don't need to call CreateTags - if we know the tag in advance, which I think is our case - but can simply pass them to the request.
I feel like it might be an easy fix for the apparent problem that we have, as we wouldn't have to take care of that, AWS would.

See the type: https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#SpotInstanceRequest

On the other hand the aws.Config that we provided is fairly simple:
https://github.com/cristim/autospotting/blob/master/core/connections.go#L22-L25

We do not set any MaxRetries, which means - according to the doc - that the services defaultMaxRetries is being used, but I couldn't get that value for sure.
I found: https://github.com/aws/aws-sdk-go/blob/master/aws/client/default_retryer.go#L23-L25 - which sounds a bit much to me; while on .net SDK it is between 4 & 10 depending on the service: http://docs.aws.amazon.com/sdk-for-net/v2/developer-guide/retries-timeouts.html

So it might be worth checking that value, and eventually increasing it if need be - as if AWS is failing even though it has a retry implemented, we might be looking at the wrong place.

cristim:
@xlr-8 this is not about the tags set on the instance (which could surely be optimized as well), but about the tags set on the spot instance request (SIR) object which identifies for which group it was created so we can attach its spot instance to the correct group.

This tagging is only attempted once, and sometimes we get a 400 because the SIR object is not yet ready to accept requests. From outside this looks like a race condition with propagation on some eventually-consistent storage.

 I think normally the SDK's retries happen only in case of API throttling, not in case of real response errors, like it happens in this case.

I started working on this and I have some code locally, but I'm struggling to properly write the tests for it since it's using time.Sleep, and I don't want the tests to sleep.

Do you have any suggestions on how to mock this?

xlr-8:
@cristim Yes, I'm aware this isn't on the instance, but the link I provided was for the SIR type, which includes tags. This is why I thought we could update the code to include the tags into the SIR rather than waiting for which-ever condition to allow us to do so.

Am I missing something out?

cristim:
Setting the SIR tags on creation would still be an improvement, apparently this should be supported by the API.

Unfortunately the instance tagging from SIR creation seems to be still unsupported, hopefully it will be available later on.

[...]

Issue #20, opened 2016-09-28
Some Spot are created without Tag


roeyazroel:
Some spot requests are fulfilled, but no 'launched-for-asg' is been assigned to this instance.

03:08:27
autoscaling.go:443: production-ecs-cluster Created spot instance request sir-dbqr5wjj
03:08:27
autoscaling.go:483: production-ecs-cluster Failed to create tags for the spot instance request InvalidSpotInstanceRequestID.NotFound: The spot instance request ID 'sir-dbqr5wjj' does not exist
03:08:28
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:28
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:33
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:33
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:38
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:38
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:43
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:43
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:48
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:48
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:53
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:53
SpotInstanceRequestId: "sir-dbqr5wjj",
03:08:58
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:08:58
SpotInstanceRequestId: "sir-dbqr5wjj",
03:09:04
autoscaling.go:335: production-ecs-cluster Refreshed details for sir-dbqr5wjj {
03:09:04
SpotInstanceRequestId: "sir-dbqr5wjj",

roeyazroel:
Maybe wrapping the following code with 'waitForRequestToFulfilled' or some timeout will work:

`func (a *autoScalingGroup) bidForSpotInstance(
	ls *ec2.RequestSpotLaunchSpecification,
	price float64) {

	svc := a.region.services.ec2

	resp, err := svc.RequestSpotInstances(&ec2.RequestSpotInstancesInput{
[...]

PR #113, opened 2017-06-25
[WIP] spotRequest: tag logic directly into the request

xlr-8:
# Bugfix Pull Request #

## Summary ##

The previous way of tagging the spot instances, was to create it, then
wait to add the various tags (those of the ASG), and the name of the ASG
to know where to attach it. However, all of this information is supposed
to be already known, so the tagging has been delegated to the AWS API.

**Important**: I'm aware that the code is currently untested, and hasn't been manually tested either - only the logic has been reviewed. Tests should follow once the logic & code have been validated working as expected. Nevertheless it does compile and passes tests as such.

Hopefully that work will fix: https://github.com/cristim/autospotting/issues/111 & https://github.com/cristim/autospotting/issues/20

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -419,7 +419,7 @@ func (a *autoScalingGroup) havingReadyToAttachSpotInstance() (*string, bool) {
 			// function timeout when waiting for the instances would break the loop,
 			// because the subsequent run would find a failed spot request instead
 			// of an open one.
-			req.waitForAndTagSpotInstance()
+			req.waitForSpotInstance()
 			activeSpotInstanceRequest = req
 		}
 
@@ -451,7 +451,7 @@ func (a *autoScalingGroup) havingReadyToAttachSpotInstance() (*string, bool) {
 				} else {
 					logger.Println(a.name, "Active bid was found, with no running "+
 						"instances, waiting for an instance to start ...")
-					req.waitForAndTagSpotInstance()
+					req.waitForSpotInstance()
 					activeSpotInstanceRequest = req
 				}
 			}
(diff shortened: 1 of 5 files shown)


xlr-8:
Yeah, I think I got tricked by the sdk-go. I thought that adding tags on spot request would propagate those to the instance, but now my understanding is that only the 'spot request' itself will be tagged with those; so it doesn't solve our problem.

I also ran into: https://github.com/aws/aws-sdk-go/issues/804 which seems to confirm my thoughts.

core/autoscaling.go
```diff
@@ -574,34 +574,28 @@ func (a *autoScalingGroup) bidForSpotInstance(
...
+	spotRequest.Tags = a.propagatedInstanceTags()
+	spotRequest.Tags = append(spotRequest.Tags, &ec2.Tag{
+		Key:   aws.String("launched-for-asg"),
+		Value: aws.String(a.name),
+	})
+	sr := spotInstanceRequest{
+		SpotInstanceRequest: spotRequest,
+		region:              a.region,
+		asg:                 a,
 	}
```
cristim: The problem is the tags set on the spotRequest.Tags are never persisted and just remain in memory, as far as I can tell that variable is set just for reading the tags, it can't be directly used for writing them back.

Because of this most of this PR is not going to work, so I'm afraid we need to hold it until the API supports tagging the SIR on creation.

cristim:
Closing this PR for now because this logic is incorrect and relies on assumptions which unfortunately are not currently implemented by the AWS EC2 API.

This should be revised once the API implements tagging of spot instance requests on creation.

PR #120, opened 2017-07-28
Add option to keep instance type

raravena80:
# Issue Type #

- Feature Pull Request

## Summary ##

This change basically adds an option `-keep_instance_type` so that when we
replace the instance in the ASG it uses the same instance type in aWS

```
$ ./autospotting -regions us-west-2 -keep_instance_type
2017/07/28 13:34:22 Starting autospotting agent, build:
2017/07/28 13:34:22 Parsed command line flags: regions='us-west-2' min_on_demand_number=0 min_on_demand_percentage=0.0 keep_instance_type=true
2017/07/28 13:34:22 main.go:97: Scanning for available AWS regions
2017/07/28 13:34:23 main.go:69: Enabled to run in us-west-2, processing region.
2017/07/28 13:34:23 region.go:66: Creating connections to the required AWS services in us-west-2
2017/07/28 13:34:23 connections.go:29: Creating Service connections in us-west-2
2017/07/28 13:34:23 connections.go:43: Created service connections in us-west-2
2017/07/28 13:34:23 region.go:70: Scanning for enabled AutoScaling groups in  us-west-2
2017/07/28 13:34:23 region.go:258: Processing page 1 of DescribeTagsPages for us-west-2
2017/07/28 13:34:23 region.go:260: us-west-2 has enabled ASG: Autospotting-ASG
2017/07/28 13:34:23 region.go:292: Processing page 1 of DescribeAutoScalingGroupsPages for us-west-2
2017/07/28 13:34:23 region.go:77: Scanning full instance information in us-west-2
2017/07/28 13:34:23 spot_price.go:24: us-west-2 Requesting spot prices
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.16xlarge in us-west-2c skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.4xlarge in us-west-2b skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.8xlarge in us-west-2b skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.16xlarge in us-west-2a skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.4xlarge in us-west-2a skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.8xlarge in us-west-2a skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.16xlarge in us-west-2b skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.4xlarge in us-west-2c skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:231: us-west-2 Instance data missing for g3.8xlarge in us-west-2c skipping because this region is currently not supported
2017/07/28 13:34:24 region.go:82: Scanning instances in us-west-2
2017/07/28 13:34:24 region.go:116: Processing page 1 of DescribeInstancesPages for us-west-2
2017/07/28 13:34:24 region.go:88: Processing enabled AutoScaling groups in us-west-2
2017/07/28 13:34:24 autoscaling.go:173: Finding spot instance requests created for Autospotting-ASG
2017/07/28 13:34:24 autoscaling.go:233: Spot instance requests were previously created for Autospotting-ASG
2017/07/28 13:34:24 autoscaling.go:245: Adding instances to Autospotting-ASG
2017/07/28 13:34:24 autoscaling.go:140: No default value for on-demand instances specified, skipping.
2017/07/28 13:34:24 autoscaling.go:747: Autospotting-ASG Counting already running on demand instances
2017/07/28 13:34:24 autoscaling.go:763: Autospotting-ASG Found 0 on-demand instances running on a total of 2
2017/07/28 13:34:24 autoscaling.go:152: Currently OnDemand running equals to the required number, skipping run
2017/07/28 13:34:24 Execution completed, nothing left to do
```


core/autoscaling.go
```diff
@@ -526,8 +531,14 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
 	}
 	logger.Println("Found on-demand instance", *baseInstance.InstanceId)
 
-	newInstanceType, err := baseInstance.getCheapestCompatibleSpotInstanceType()
-
+	// Check option to keep instance type
+	// If we keep the instance type we don't need to calculate the
+	// compatible instance type.
+	if a.region.conf.KeepInstanceType {
+		newInstanceType = *baseInstance.InstanceId
```
cristim: Should't this be `*baseInstance.InstanceType` instead?

autospotting.go
```diff
@@ -98,6 +105,9 @@ func (c *cfgData) parseCommandLineFlags() {
 			autospotting.OnDemandPercentageLong+
 			"\n\tIt is ignored if min_on_demand_number is also set.")
 
+	flag.BoolVar(&c.KeepInstanceType, "keep_instance_type", false,
+		"If specified the instance type of the on-demains instances "+
+			"is matched")
```
cristim: I'd use "If specified, the spot instances will have the same instance type as the initial on-demand instances", I think it's a bit more clear

autospotting.go
```diff
@@ -98,6 +105,9 @@ func (c *cfgData) parseCommandLineFlags() {
 			autospotting.OnDemandPercentageLong+
 			"\n\tIt is ignored if min_on_demand_number is also set.")
 
+	flag.BoolVar(&c.KeepInstanceType, "keep_instance_type", false,
+		"If specified, the spot instances will have the same instance type "+
+			"as the initial on-demand instances")
```
xlr-8: Why not making it a string, so that any type can be given?
A special type such as `current` could be given to indicate to take the same type as the one in the ASGs, what do you think?

core/autoscaling.go
```diff
@@ -334,6 +335,47 @@ func (a *autoScalingGroup) replaceOnDemandInstanceWithSpot(
...
+	output, err := svc.DescribeTags(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case autoscaling.ErrCodeInvalidNextToken:
+				logger.Println(autoscaling.ErrCodeInvalidNextToken, aerr.Error())
+			case autoscaling.ErrCodeResourceContentionFault:
+				logger.Println(autoscaling.ErrCodeResourceContentionFault, aerr.Error())
+			default:
+				logger.Println(aerr.Error())
```
xlr-8: Is the switch case really needed? Could we always print `aerr.Code` and `aerr.Error()` in order to remove it?

core/autoscaling.go
```diff
@@ -334,6 +335,47 @@ func (a *autoScalingGroup) replaceOnDemandInstanceWithSpot(
...
+
+	instanceTypes, err := a.region.requestSpotInstanceTypes()
+	if err != nil {
+		logger.Println(err.Error())
+		return "", err
+	}
+	for _, it := range instanceTypes {
+		if *output.Tags[0].Value == it {
+			return *output.Tags[0].Value, nil
+		}
```
xlr-8: This loop is made because some on-demand instance types do not exist as spot instance, is it correct?
Otherwise, I'm not sure to understand it.

core/region.go
```diff
@@ -239,6 +239,27 @@ func (r *region) requestSpotPrices() error {
...
+	var instTypes []string
+
+	s := spotPrices{conn: r.services}
+
+	// Retrieve all current spot prices from the current region.
+	// TODO: add support for other OSes
+	err := s.fetch("Linux/UNIX", 0, nil, nil)
+
+	if err != nil {
+		return nil, errors.New("Couldn't fetch spot prices in" + r.name)
```
xlr-8: I believe there is a missing space in the error message right after the `in`

core/autoscaling.go
```diff
@@ -526,12 +573,27 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
...
+	// Check option to keep instance type
+	// If we keep the instance type we don't need to calculate the
+	// compatible instance type.
+	// Also ignore if it's already specified in the instance-type tag
+	if newInstanceType == "" {
+		if a.region.conf.KeepInstanceType {
+			newInstanceType = *baseInstance.InstanceType
+		} else {
+			newInstanceType, err = baseInstance.getCheapestCompatibleSpotInstanceType()
+		}
```
xlr-8: I think that having the option as a `string` rather than a `bool` would also potentially simplify this `if` / `else` as the type would be already known completely or not at all, thus removing the possibility of a temporary `""` type.

core/autoscaling_test.go
```diff
@@ -2674,3 +2674,68 @@ func TestReplaceOnDemandInstanceWithSpot(t *testing.T) {
 		})
 	}
 }
+
+func TestGetInstanceTypeByTagInASG(t *testing.T) {
+	tests := []struct {
+		name     string
+		asgTags  []*autoscaling.TagDescription
+		tagKey   string
+		expected *string
```
xlr-8: The prototype of `getInstanceTypeByTagInASG` is : `func (a *autoScalingGroup) getInstanceTypeByTagInASG() (string, error) {` so I assume that the `error` field should also be tested. #

core/region.go
```diff
@@ -239,6 +239,27 @@ func (r *region) requestSpotPrices() error {
 	return nil
 }
 
+func (r *region) requestSpotInstanceTypes() ([]string, error) {
```
xlr-8: This method doesn't appear to be tested ?

core/autoscaling.go
raravena80: Yes and also for validation.

core/autoscaling.go
raravena80: ok I'll take a look

core/region.go
raravena80: I just added a test on the latest commit

core/region.go
raravena80: Good catch. I'll add the space

autospotting.go
raravena80: good idea. Let me change it

core/autoscaling.go
raravena80: I can simplify. There's no need to break down based on `aerr.code()`

raravena80:
@cristim @xlr-8 I've made the requested changes. Can you take a look. Thx

autospotting.go
```diff
@@ -30,8 +30,15 @@ func main() {
 func run() {
 	log.Println("Starting autospotting agent, build:", Version)
 
-	log.Printf("Parsed command line flags: regions='%s' min_on_demand_number=%d min_on_demand_percentage=%.1f",
-		conf.Regions, conf.MinOnDemandNumber, conf.MinOnDemandPercentage)
+	log.Printf("Parsed command line flags: "+
+		"regions='%s' "+
+		"min_on_demand_number=%d "+
+		"min_on_demand_percentage=%.1f "+
+		"keep_instance_type=%t",
```
cristim: I think with the new semantic a better name for this parameter would be something like "allowed_instance_types"

autospotting.go
```diff
@@ -98,6 +105,10 @@ func (c *cfgData) parseCommandLineFlags() {
 			autospotting.OnDemandPercentageLong+
 			"\n\tIt is ignored if min_on_demand_number is also set.")
 
+	flag.StringVar(&c.KeepInstanceType, "keep_instance_type", "",
+		"If specified, the spot instances will have a specific instance type:\n"+
+			"\tcurrent: the same as iniial on-demand instances\n"+
```
cristim: Typo: should be "initial"

core/autoscaling.go
```diff
@@ -334,6 +335,40 @@ func (a *autoScalingGroup) replaceOnDemandInstanceWithSpot(
 	return a.detachAndTerminateOnDemandInstance(odInst.InstanceId)
 }
 
+func (a *autoScalingGroup) getInstanceTypeByTagInASG() (string, error) {
```
cristim: I don't understand what this function does and why we have it.

I think the tag is already available in the Tags attribute of each group, so we don't need to query it again.

core/autoscaling.go
```diff
@@ -526,12 +566,29 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
...
+
+	// Check option to keep instance type
+	// If we keep the instance type we don't need to calculate the
+	// compatible instance type.
+	// Also ignore if it's already specified in the instance-type tag
+	if newInstanceType == "" {
+		if a.region.conf.KeepInstanceType == "current" {
+			newInstanceType = *baseInstance.InstanceType
+		} else if a.region.conf.KeepInstanceType != "" {
+			newInstanceType = a.region.conf.KeepInstanceType
```
cristim: I would convert this logic into a match-based filter, allowing the user to pass multiple allowed instance types in a comma-separated list, and should also support using wildcards:

allowed_instance_types: c3.*, m4.*
or 
allowed_instance_types: m4.large
or
allowed_instance_types: same (or current)

This is much like we already do for the filter of allowed regions.

raravena80: that's a different behavior of what we want meaning that's like another feature. 
Do we need to implement that as part of the PR?

core/autoscaling.go
raravena80: You think? or it is ?  if it is, wish you would have caught this earlier, I spent sometime creating a test for this function.  Let me know.

autospotting.go
raravena80: Noted. thx

autospotting.go
raravena80: This is a different feature.  We want to keep the instance type 'exactly' of what it is. Do we need to implement in this PR? I believe we should open another PR, feature for that.

core/autoscaling.go
raravena80: hmm.. come to think of it.  I think this works for us as along as we specify a single instance on allowed instance types.

core/autoscaling.go
cristim: Sorry for wasting your time, I missed it in the previous review.

I was referring to this: https://docs.aws.amazon.com/sdk-for-go/api/service/autoscaling/#Group, notice that it has a Tags field which should be already populated.

The other functionality using configurations set on the group already takes the tags from there, have a look at https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L76 and https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L756

raravena80: Sounds good, I'll take a look thx

core/autoscaling.go
cristim: Exactly, I would prefer to implement that from the start, what you need is just a particular case of it and should need minor changes on top of what you already have done.

But if you're only willing to scratch your own itch, I understand and that's okay, we can enhance it to do  that later. It's already great that you are laying down a lot of the foundations for that extension and I can't thank you enough for this :)

autospotting.go
cristim: I'm fine either way, but that should be just a few more lines on top of what you wrote already.

core/autoscaling.go
```diff
@@ -566,29 +533,33 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
...
+		allowed = allowedInstanceTypes
+		allowedList = strings.Split(allowed, ", ")
+	}
+
+	if allowed == "current" {
+		newInstanceType = *baseInstance.InstanceType
+	} else if allowed != "" {
+		newInstanceType = allowed
+	} else {
+		newInstanceType, err = baseInstance.getCheapestCompatibleSpotInstanceType(allowedList)
```
cristim: I would wrap this configuration handling logic inside a function returning the allowedList

raravena80: hmm, not sure I understand. 
Meaning `newInstanceType` becomes allowedList or the function returns `newInstanceType`?

core/autoscaling.go
```diff
@@ -566,29 +533,33 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
...
-			logger.Println("No cheaper compatible instance type was found, "+
-				"nothing to do here...", err)
-			return errors.New("no cheaper spot instance found")
-		}
+	// Also ignore if it's already specified in the allowed-instance-types tag
+	allowedInstanceTypesTag := strings.Replace(*a.getTagValue("allowed-instance-types"), " ", ",", -1)
+	allowedInstanceTypes := strings.Replace(a.region.conf.AllowedInstanceTypes, " ", ",", -1)
+
+	// Tag has priority
+	if allowedInstanceTypes == "" {
```
raravena80: Cool let me change it

core/autoscaling.go
cristim: I thought about introducing a function that contains all that configuration logic, say allowedInstanceTypes() and call it something like
``` newInstanceType, err = baseInstance.getCheapestCompatibleSpotInstanceType(a.allowedInstanceTypes())```

Comment (raravena80) [inline core/autoscaling.go]: Ah cool, sounds good.

File: core/autoscaling.go [hunk 16]
```diff
@@ -508,13 +507,12 @@ func (a *autoScalingGroup) havingReadyToAttachSpotInstance() (*string, bool) {
 	return spotInstanceID, false
 }
 
-func (a *autoScalingGroup) allowedInstanceTypes() (string, error) {
+func (a *autoScalingGroup) getAllowedInstanceTypes(baseInstance *instance) string {
```
cristim: I would return a slice of strings so we don't have to process the result afterwards with `strings.Split`

raravena80: @cristim changed.

core/region.go
cristim: I will dismiss this review rejection since the tests were implemented.

cristim:
@raravena80 you finally made it, huge thanks for your contribution!

I really hope the lengthy process won't discourage you from doing further contributions.

stevenpall:
To be clear, how should this option be used with an existing installation? Will it be a CloudFormation parameter going forward? Perhaps some documentation should be added around this.

cristim:
Yeah, we missed the documentation this time, we should add it as part of a checklist to the contribution [guidelines](https://github.com/cristim/autospotting/blob/master/CONTRIBUTING.md).

@stevenpall you just need to update the sutospotting build using CloudFormation or Terraform, as per the normal upgrade procedure documented [here](https://github.com/cristim/autospotting/blob/master/START.md#updates-and-downgrades), then tag the AutoScaling group with the newly introduced tags, which also need to be documented as requested on #121.

The tag is named `allowed_instance_types` and accepts a comma(or whitespace)-separated list of instance types.

The AutoSpotting CloudFormaiton stack should also be extended to accept this as an argument, but we're not there yet.

stevenpall:
Perfect, thanks. I could probably add some documentation around this in a separate PR if you would like?

cristim:
@stevenpall that would be awesome, thanks!

stevenpall:
Just tested this, and it does not appear to be working as expected. I'm running the latest build, `dv/lambda_build_380.zip` and have tagged the ASG with `allowed_instance_types` set to `c4.large`. When the script has an on-demand instance to replace, it appears to still be choosing from a variety of instance types and does not limit the ultimate selection to the specified instance type.

One thing I noticed in the logs that I'm not sure about is this:
```
Parsed command line flags: regions='us-east-1' min_on_demand_number=0 min_on_demand_percentage=0.0 allowed_instance_types=%!t(string=)
```
The `allowed_instance_types=%!t(string=)` bit looks odd to me. Is that alright?

raravena80:
@stevenpall Something is odd.  Did you build it yourself ? or where did you get the build ?

raravena80:
So there's an issue with the debug message `allowed_instance_types=%!t(string=)` . However, in essence it means that allowed_instance_types is empty

stevenpall:
@raravena80 This is from the normal S3 endpoint (http://s3.amazonaws.com/cloudprowess). Do any flags need to be passed to the binary for this to work or should it work out of the box after adding the above-stated ASG tags?

cristim:
That's normal when just setting it in the Autoscaling group tag. This message is just about the global configuration of the Lambda function as set up via Cloudformation, since the stack wasn't yet changed to support this parameter.

cristim:
@raravena80 did you test this code before submitting? I just did the code review and didn't get to actually try this feature yet.

raravena80:
@cristim haven't had the chance to try all cases yet.  I tested from the command line, not from Lambda and not all the different combinations.  Although I did add a few go tests.

cristim:
Okay, that explains it. I will add this requirement to the contributions checklist.

@stevenpall for now try to add it to the Lambda function environment variables, that should do the trick with the caveat that it will apply across all the groups managed by autospotting

cristim:
@stevenpall are you still stuck with this?

I'd like to get this issue sorted out sometime these days.

stevenpall:
@cristim I figured I should wait till we get the code working before writing the documentation (in case something changes).

cristim:
@stevenpall sure, the important thing is to get it work first, the documentation is secondary.

Have you got any further, or do you need any help?

PR #127, opened 2017-08-16
Also use GPU number for instance compatibility

### LINKED ISSUE #126: Older GPU instance types get replaced with non-GPU instances
The tool works great for all instance types except the p2 family. These seem to be replaced with regular instances that have the same or higher number of CPUs and amount of RAM. Is there any configuration that needs to happen to support GPU spot instances?

cristim:
Fixes #126.

# Issue Type #

- Feature Pull Request

## Summary ##

This should fix #126

### DIFF
--- a/core/instance.go
+++ b/core/instance.go
@@ -99,6 +99,7 @@ type instance struct {
 type instanceTypeInformation struct {
 	instanceType             string
 	vCPU                     int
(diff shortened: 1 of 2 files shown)


cristim:
Thanks @artemnikitin for the speedy review!

PR #132, opened 2017-08-27
Use Go 1.9 in TravisCI

cristim:
# Issue Type

- Feature Pull Request

## Summary

Update TravisCI builds to the latest golang version, 1.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-- 1.8
+- 1.9
 
 sudo: required
 


xlr-8:
You could also make a list as `go` is an array parameter, so you could specify:
- 1.9
- master
Which according to the doc should run 2 builds for 1.9 and the latest version from master.
Ref: https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use

cristim:
We're depending on the Go version shipped by https://github.com/eawsy/aws-lambda-go-shim, which at the moment is 1.9, since that's what's used for the official production builds.

The goal was to use the same version in Travis also for fetching the dependencies before starting the build.

Fortunately Go is legendary when it comes to backwards compatibility so we should be fine anyway.

PR #134, opened 2017-08-28
Minor fix for the options output message.

raravena80:
# Issue Type

- Bugfix Pull Request

## Summary

This is a very minor fix for the options parsing. It's a leftover from https://github.com/cristim/autospotting/pull/120

   to it.
   and
   configurations.

### DIFF
--- a/autospotting.go
+++ b/autospotting.go
@@ -34,7 +34,7 @@ func run() {
 		"regions='%s' "+
 		"min_on_demand_number=%d "+
 		"min_on_demand_percentage=%.1f "+
-		"allowed_instance_types=%t",
+		"allowed_instance_types=%v",
 		conf.Regions,
 		conf.MinOnDemandNumber,
 		conf.MinOnDemandPercentage,


raravena80:
@cristim looks like the build is failing because of 
```
/usr/local/go/src/github.com/cristim/ec2-instances-info
```

cristim:
@raravena80 that shouldn't cause this, I suspect it's the latest change in https://github.com/eawsy/aws-lambda-go-shim which I think changed some things about handling of the GOPATH variable.

@fsenart, does this look familiar? https://travis-ci.org/cristim/autospotting/builds/269340055?utm_source=github_status&utm_medium=notification

fsenart:
@cristim in order to be compliant with the new Go policy which sets a default GOPATH to $HOME/go and to be able to handle multiple GOPATHs, we've removed the default GOPATH=/go from the docker image. Also your build will pass if you add `-e GOPATH=/go` to the `docker run` command. Sorry for the inconvenience.

cristim:
Thanks @fsenart, that did the trick!

cristim:
@raravena80 let's abandon this, I already applied this change in #135

PR #135, opened 2017-08-28
Fix TravisCI builds

cristim:
# Issue Type

- Bugfix Pull Request

## Summary

Fix TravisCI build after latest Golang Lambda shim container updates


cristim:
@raravena80 this should fix it

raravena80:
👍

PR #142, opened 2017-09-08
simplify termination logic

vecchp:
# Issue Type

- Feature Pull Request

## Summary

Implements #141 

   to it.
   and
   configurations.


vecchp:
Just saw this was already done

cristim:
@vecchp please also help us review #141, we need to help testing this work.

PR #158, opened 2017-11-07
Update README.md

russellballestrini:
We use autospotting at [Remind](https://www.remind.com) and might be one of the bigger users as far as the amount of spot instances we managed using autospotting.


cristim:
Thanks Russell, as far as I know Remind is indeed one of the largest users, glad to have you guys on the top of the users list.

PR #163, opened 2017-11-28
add new global option on_demand_price_multiplier to support volume discounts

masneyb:
# Issue Type

- Feature Pull Request

## Summary

This patch adds a new global option called on_demand_price_multiplier that is primarily intended to be used to support EC2 volume discounts. The option could also be used to set your set your bid price to be higher than the on demand price to reduce the chances that your spot instances will be terminated.

Note: There are currently no tests associated with this change. I started to write some tests for region.determineInstanceTypeInformation(), however ec2instancesinfo.regionPrices is not exposed publicly. Before I started sending you other pull requests to make these changes I first wanted to see if you would accept a patch like this or if you would like this feature to be implemented in a different manner.

### DIFF
--- a/START.md
+++ b/START.md
@@ -217,9 +217,10 @@ Usage of ./autospotting:
         Can be overridden on a per-group basis using the tag autospotting_on_demand_percentage
         It is ignored if min_on_demand_number is also set.
   -on_demand_price_multiplier float
-        Multiplier for the on-demand price. This is useful for volume discounts or if you want
-        to set your bid price to be higher than the on demand price to reduce the chances that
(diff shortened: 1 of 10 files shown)


cristim:
I'll keep this open for a few more days to give @masneyb some time to write some tests for this.

masneyb:
I added unit tests for this option. This pull request https://github.com/cristim/ec2-instances-info/pull/4 is required to be merged first in order for these tests to compile.

cristim:
Reopened to trigger a new Travis build

cristim:
Huge thanks for your contribution, looking forward for more 😁

Issue #104, opened 2017-04-02
Autoscale Down temporarily causes less than minimum on-demand number and may cause downtime on small groups


ahaverbuch:
# Github issue #

## Issue type ##

Feature Idea

## Summary ##

We are seeing an issue on autoscaling groups that are  autoscaling while autospotting is running.

Two fixes would help - this covers one of them (#103 is the other)
1) If you have an environment, say like this: 8 machines, min of 4 on-demand, oldest-instance for scale down policy. Autospotting runs, you get 4 on-demand (older), 4-spot (which are newer, as they replace other on-demand). If you then scale down, say to 6, you end up with only 2 on-demand, and 4-spot, which is smaller than the number you intend. It would be nice to have support to bring back to on-demand to get back to the intended level in order to minimize risk of spot market volatility killing a larger % of machines than intended to be running.

cristim:
I guess we could terminate a spot instance if the ratio of spot vs on-demand is more than configured, and let autoscaling bring up a new on demand instance in its place

xlr-8:
Correct me if I'm wrong, but if you specify `-min_on_demand_number=4`, shouldn't this behave as you expect? Or are you making a point about ratio between on-demand/spot instance as @cristim suggested?

That's what this function is intended for: https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L145, if you end-up in a situation with more spot instances, than on-demand required, then a spot instance gets terminated in order to bring back progressively this requirement.

ahaverbuch:
That's good to know. That function does cause the behavior of getting back to on demand ratio eventually. But it actually itself can cause an issue as it's not graceful; if you are running close to the minimum number of machines you need to handle traffic, the terminate random spot will bring you under at least until the on demand machine comes up

If instead it did the opposite of the spot bring up, by bringing up an on-demand outside of the asg, then attaching it after grace period, then detaching spot, you wouldn't fall under the number of machines intending to be running.

cristim:
Yes, it's not graceful, but remember that you had just scaled down, which means you are likely to have more than enough capacity in the group so losing one more spot instance(which may anyway happen at any time due to market price fluctuations) and waiting for AutoScaling to replace it with a new on-demand node shouldn't cause much problems.

Have you seen this causing problems in your environment?

ahaverbuch:
I agree this can happen at any time due to spot, but there's scenarios where this forces an error state where it doesn't need to.

I see two items:

[...]

PR #164, opened 2017-11-29
prevent termination of last instance, should improve #104

universam1:
# Issue Type

- Bugfix Pull Request

## Summary

Prevent actively destruction of the last remaining instance. This is not intended to be an ultimate solution for #104 but should prevent harm by itself.

Open for any suggestion as I am new to golang.

### DIFF
--- a/core/autoscaling_test.go
+++ b/core/autoscaling_test.go
@@ -1086,6 +1086,22 @@ func TestNeedReplaceOnDemandInstances(t *testing.T) {
 			desiredCapacity: aws.Int64(0),
 			expectedRun:     false,
 		},
+		{name: "ASG has only one remaining instance, less than enough on-demand",
+			asgInstances: makeInstancesWithCatalog(
+				map[string]*instance{
+					"id-1": {
+						Instance: &ec2.Instance{
+							State:             &ec2.InstanceState{Name: aws.String("running")},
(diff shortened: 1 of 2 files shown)


cristim:
@universam1 Thanks for your contribution to the project!

Please create some tests for the function you touched with this PR. It may not have any yet, but we're trying to add tests as we go even for existing code.

core/autoscaling_test.go
```diff
@@ -1086,6 +1086,22 @@ func TestNeedReplaceOnDemandInstances(t *testing.T) {
...
+						Instance: &ec2.Instance{
+							State:             &ec2.InstanceState{Name: aws.String("running")},
+							Placement:         &ec2.Placement{AvailabilityZone: aws.String("eu-west-1a")},
+							InstanceLifecycle: aws.String("spot"),
+						},
+					},
+				},
+			),
+			minOnDemand:     1,
+			desiredCapacity: aws.Int64(0),
```
cristim: Shouldn't we also have a case for min=desired=1?

universam1:
added a test to cover my piece of code. At least a little positive influence to coverage.
Please don't expect high level contribution as I started to deal with go yet since yesterday.
However, I would see this PR as a quick fix rather than a major refactor.

core/autoscaling_test.go
universam1: Indeed, thanks for check

Issue #119, opened 2017-07-25
Implement multiple spot bidding strategies


cristim:
# Feature idea #

Add support for multiple bidding strategies, configurable on a per-group level.

## Detailed explanation ##

At the moment the only supported strategy is hardcoded to bid the original on-demand instance's hourly price for any spot instances added to the group.

This is quite safe and is suitable for production use where longer-running instances are desirable, but it fails to handle the spot price increases that don't go over the on-demand price but are significantly higher than when the instance was launched, and may be not the most cost-efficient strategy.

Alternate strategies may be to place the currently spot price plus a configurable margin, such as 10%, which would capture most significant spot price increases.

An absolute margin may also be desirable, allowing to place bids a fraction of a cent over the current price, which would aggressively optimize for cost savings since the last runtime hour before spot termination is free of charge.

For best results this may need to be coupled with the spot-to-spot replacement to be implemented in #118, in order to avoid the cost incurred by running temporary on-demand instances.

briantyr:
Good write up - I can see this being a long-term goal to become much more attractive to CIO's / those that "pay the bills" to be able to configure multiple bidding strategies.  Good catch noting all that down in the description @cristim

cristim:
@briantyr please have a look, we now have an initial implementation of this feature.

Feel free to give it a try and it would be great if you could give us some feedback about this.

deinspanjer:
Very exciting!  It will be a couple of weeks before I'll be able to actually test it out, but I'll read over the patch and any implementation doc and provide feedback where and as soon as I can.

PR #175, opened 2017-12-03
Added SpotInstance Percentage Multiplier with dynamic tags

kartik894:
# Issue Type

- Feature Pull Request, implements #119.

## Summary

Added configurable policy to place bids at a multiple of spot instance price or at on-demand price. Also added dynamic tag to enable this feature for an auto-scaling group.

Code review process:

The issue should be tagged with 'review wanted' label before the checklist below
is satisfied from the perspective of the PR author and the code is ready to be
peer-reviewed.

   to it.
   and
   configurations.

### DIFF
--- a/START.md
+++ b/START.md
@@ -125,6 +125,8 @@ module "autospotting" {
   autospotting_min_on_demand_percentage = "50.0"
   autospotting_regions_enabled = "eu*,us*"
   on_demand_price_multiplier = "1.0"
+  spot_price_buffer_percentage = "10.0"
+  bidding_policy = "normal"
 
   lambda_zipname = "./lambda.zip"
   lambda_runtime = "python2.7"
@@ -224,6 +226,17 @@ Usage of ./autospotting:
         or if you want to set your bid price to be higher than the on demand
         price to reduce the chances that your spot instances will be terminated.
         (default 1)
+  -spot_price_buffer_percentage float
+        Percentage Value of the bid above the current spot price. A spot
+        bid would be placed at a value
+        current_spot_price * [1 + (spot_price_buffer_percentage/100.0)] .
+        The main benefit is that it protects the group from running spot instances
+        that got significantly more expensive than when they were initially launched,
+        but still somewhat less than the on-demand price. (default 10.0)
+  -bidding_policy string
+        Policy choice for spot bid. If set to 'normal', we bid at the
+        on-demand price. If set to 'aggressive', we bid at a multiple of
+        the spot price. (default "normal")
   -regions="": Regions where it should be activated (comma or whitespace separated
         list, also supports globs), by default it runs on all regions.
         Example: ./autospotting -regions 'eu-*,us-east-1'
--- a/autospotting.go
+++ b/autospotting.go
@@ -42,12 +42,16 @@ func run() {
 		"min_on_demand_number=%d "+
 		"min_on_demand_percentage=%.1f "+
 		"allowed_instance_types=%v "+
-		"on_demand_price_multiplier=%.2f",
+		"on_demand_price_multiplier=%.2f"+
+		"spot_price_buffer_percentage=%.3f"+
+		"bidding_policy=%s",
 		conf.Regions,
 		conf.MinOnDemandNumber,
 		conf.MinOnDemandPercentage,
 		conf.AllowedInstanceTypes,
-		conf.OnDemandPriceMultiplier)
+		conf.OnDemandPriceMultiplier,
+		conf.SpotPriceBufferPercentage,
+		conf.BiddingPolicy)
 
 	autospotting.Run(conf.Config)
 	log.Println("Execution completed, nothing left to do")
@@ -123,6 +127,18 @@ func (c *cfgData) parseCommandLineFlags() {
 		"Multiplier for the on-demand price. This is useful for volume discounts or if you want to\n"+
 			"\tset your bid price to be higher than the on demand price to reduce the chances that your\n"+
 			"\tspot instances will be terminated.")
+
+	flag.Float64Var(&c.SpotPriceBufferPercentage, "spot_price_buffer_percentage", 10,
+		"Percentage Value of the bid above the current spot price. A spot bid would be placed at a value :\n"+
+			"\tcurrent_spot_price * [1 + (spot_price_buffer_percentage/100.0)]. The main benefit is that\n"+
+			"\tit protects the group from running spot instances that got significantly more expensive than\n"+
+			"\twhen they were initially launched, but still somewhat less than the on-demand price. Can be\n"+
+			"\tenforced using the tag: "+autospotting.SpotPriceBufferPercentageTag+". If the bid exceeds\n"+
+			"\tthe on-demand price, we place a bid at on-demand price itself.")
+
+	flag.StringVar(&c.BiddingPolicy, "bidding_policy", "normal",
+		"Policy choice for spot bid. If set to 'normal', we bid at the on-demand price. If set to 'aggressive',\n"+
+			"\twe bid at a percentage value above the spot price. ")
 
 	v := flag.Bool("version", false, "Print version number and exit.")
 
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -2,14 +2,13 @@ package autospotting
 
 import (
 	"errors"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"math"
 	"strconv"
 	"strings"
 	"time"
-
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/autoscaling"
-	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
 const (
@@ -23,9 +22,24 @@ const (
 	// absolute number.
 	OnDemandNumberLong = "autospotting_on_demand_number"
 
+	// BiddingPolicyTag stores the bidding policy for the spot instance
+	BiddingPolicyTag = "autospotting_bidding_policy"
+
+	// SpotPriceBufferPercentageTag stores percentage value above the
+	// current spot price to place the bid
+	SpotPriceBufferPercentageTag = "spot_price_buffer_percentage"
+
 	// DefaultMinOnDemandValue stores the default on-demand capacity to be kept
 	// running in a group managed by autospotting.
 	DefaultMinOnDemandValue = 0
+
+	// DefaultSpotPriceBufferPercentage stores the default percentage value
+	// above the current spot price to place a bid
+	DefaultSpotPriceBufferPercentage = 10.0
+
+	// DefaultBiddingPolicy stores the default bidding policy for
+	// the spot bid on a per-group level
+	DefaultBiddingPolicy = "normal"
 )
 
 type autoScalingGroup struct {
@@ -60,6 +74,21 @@ func (a *autoScalingGroup) loadPercentageOnDemand(tagValue *string) (int64, bool
 	return DefaultMinOnDemandValue, false
 }
 
+func (a *autoScalingGroup) loadSpotPriceBufferPercentage(tagValue *string) (float64, bool) {
+	spotPriceBufferPercentage, err := strconv.ParseFloat(*tagValue, 64)
+
+	if err != nil {
+		logger.Printf("Error with ParseFloat: %s\n", err.Error())
+		return DefaultSpotPriceBufferPercentage, false
+	} else if spotPriceBufferPercentage <= 0 {
+		logger.Printf("Ignoring out of range value : %f\n", spotPriceBufferPercentage)
+		return DefaultSpotPriceBufferPercentage, false
+	}
+
+	logger.Printf("Loaded SpotPriceBufferPercentage value to %f from tag %s\n", spotPriceBufferPercentage, SpotPriceBufferPercentageTag)
+	return spotPriceBufferPercentage, true
+}
+
 func (a *autoScalingGroup) loadNumberOnDemand(tagValue *string) (int64, bool) {
 	onDemand, err := strconv.Atoi(*tagValue)
 	if err != nil {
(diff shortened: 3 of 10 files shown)


START.md
```diff
@@ -221,6 +223,18 @@ Usage of ./autospotting:
         or if you want to set your bid price to be higher than the on demand
         price to reduce the chances that your spot instances will be terminated.
         (default 1)
+  -spot_price_multiplier float
+        Multiplier for the spot price. Placing bids a fraction of a cent over the
+        current price would aggressively optimize for cost savings since the last
+        runtime hour before spot termination is free of charge.
+        (default 1.1)
+
+  -spot_price_multiplier_flag bool
```
cristim: I would prefer to have this a multi-option flag instead of a boolean, perhaps named something like 'bidding_policy', and accepting string values such as 'normal' and 'aggressive' and document them in detail. This way we can add more policies later on.

autospotting.go
```diff
@@ -30,7 +30,8 @@ func main() {
 func run() {
 	log.Println("Starting autospotting agent, build ", Version, "expiring on", ExpirationDate)
 
-	if isExpired(ExpirationDate) {
+	if false {
```
cristim: Please don't change the expiration logic.

core/autoscaling.go
```diff
@@ -182,6 +239,8 @@ func (a *autoScalingGroup) process() {
 	a.scanInstances()
 	a.loadDefaultConfig()
 	a.loadConfigFromTags()
+	a.loadDefaultSpotConfig()
+	a.loadSpotConfigFromTags()
```
xlr-8: I think those 2 methods could be simplified (by removing the boolean) and put directly in the previous 2 methods (`loadDefaultConfig` & `loadConfigFromTags`) that were created for that; in case extra elements were to be added.

While re-reading the code I also realised it could be slightly re-arranged in order to load the default configuration and ASG configuration earlier than during the processing, see https://github.com/cristim/autospotting/issues/180.

core/autoscaling.go
```diff
@@ -88,6 +98,36 @@ func (a *autoScalingGroup) loadConfOnDemand() bool {
...
+	if err != nil {
+		return DefaultSpotPriceMultiplierFlag, true
+	}
+
+	logger.Printf("Loaded SpotPriceMultiplierFlag value with %t from tag %s\n", spot, SpotBidChoice)
+	return spot, true
+}
+
+func (a *autoScalingGroup) loadConfSpot() bool {
+	tagList := [1]string{SpotBidChoice}
```
xlr-8: Why declaring an array of string of 1 and not directly a string `tagKey`?

core/autoscaling.go
```diff
@@ -536,6 +595,18 @@ func (a *autoScalingGroup) getAllowedInstanceTypes(baseInstance *instance) []str
...
+
+	if a.SpotPriceMultiplierFlag == true {
+		logger.Println("Launching spot instance with a bid =", currentSpotPrice*a.region.conf.SpotPriceMultiplier)
+		return currentSpotPrice * a.region.conf.SpotPriceMultiplier
+	}
+
+	logger.Println("Launching spot instance with a bid =", baseOnDemandPrice)
+	return baseOnDemandPrice
+}
+
```
xlr-8: This method would be obsolete if the default parameter was added and the flag removed. The multiplier could then either be used here: https://github.com/cristim/autospotting/blob/master/core/region.go#L236
or here: https://github.com/cristim/autospotting/blob/master/core/instance.go#L160-L162

Just like this new parameter actually: https://github.com/cristim/autospotting/pull/163/files#diff-17a0faaa6296d309bb099b250ff6ead1R162.

I'd tend to prefer putting them in the `instance.go` file as it seems more logical, but the main method would need to take new parameters such as `biding_policies` or `multipliers` or something like this.

@cristim what are your thoughts on this?

core/autoscaling.go
```diff
@@ -584,7 +654,7 @@ func (a *autoScalingGroup) launchCheapestSpotInstance(
 		*azToLaunchIn)
 
 	logger.Println("Bidding for spot instance for ", a.name)
-	return a.bidForSpotInstance(spotLS, baseOnDemandPrice)
+	return a.bidForSpotInstance(spotLS, a.getPricetoBid(baseOnDemandPrice, currentSpotPrice))
```
xlr-8: See previous comment

core/autoscaling_test.go
```diff
@@ -2808,3 +2970,64 @@ func TestGetAllowedInstaceTypes(t *testing.T) {
...
+		// newInstance := asg.region.instanceTypeInformation["t2.medium"]
+		currentSpotPrice := tt.currentSpotPrice
+		currentOnDemandPrice := tt.currentOnDemandPrice
+		actualPrice := asg.getPricetoBid(currentOnDemandPrice, currentSpotPrice)
+		if math.Abs(actualPrice-tt.want) > 0.000001 {
+			t.Errorf("multiplier = %.2f, flag = %t, expected price = %.5f, want %.5f, currentSpotPrice = %.5f",
+				tt.multiplier, tt.flag, actualPrice, tt.want, currentSpotPrice)
+		}
+	}
+}
```
xlr-8: Tests not yet reviewed due to previous comments, please let me know if you'd like me to do so anyway

cristim:
@xlr-8 we can't entirely get rid of the boolean or the string flag because this is meant to introduce an additional bidding policy, so we basically need a switch to choose between the old and this one and a second piece of configuration (or maybe later more) for tweaking this additional bidding strategy.

In theory we could swing to the new bidding strategy only if its configuration tweaking option is different from a default value, but I would prefer to use an explicit, dedicated flag for this switch.

xlr-8:
Alright, thx @cristim I'll update my comments accordingly, sorry for the delay @kartik894

START.md
xlr-8: The idea would be to have a "policy type" and a tag associated with the configuration of this specific policy, is that correct @cristim ?

core/autoscaling_test.go
```diff
@@ -686,6 +686,168 @@ func TestLoadConfigFromTags(t *testing.T) {
...
+	for _, tt := range tests {
+		a := autoScalingGroup{Group: &autoscaling.Group{}}
+		value, _ := a.loadSpotPriceMultiplierFlag(tt.tagValue)
+
+		if value != tt.valueExpected {
+			t.Errorf("LoadSpotPriceMultiplierFlag returned: %t, expected: %t", value, tt.valueExpected)
+		}
+
+	}
+}
```
xlr-8: Would need to be updated to fetch various decided policy names

START.md
kartik894: Yes

autospotting.go
kartik894: Will change this.

core/autoscaling.go
kartik894: It can be done, but at the repo level - requestSpotPrices() might be used for another bidding policy in the future and I wouldn't want to touch that. isPriceCompatible() is used to determine the choice of spot instances to be spawned for the current on-demand instance - Also wouldn't want to change that. I did think of these scenarios before I begun the implementation, but I figured it would be best placed in autoscaling.go since that is where you call for the bidPrice to be placed.

core/autoscaling.go
kartik894: loadDefaultConfig has dedicated tests and merging mine with that will be a bit messy. What do you think @cristim ?

cristim: I  agree with @xlr-8, the purpose of tests is for making such refactoring easier, let's merge these because they do more or less the same thing.

START.md
```diff
@@ -221,6 +223,18 @@ Usage of ./autospotting:
         or if you want to set your bid price to be higher than the on demand
         price to reduce the chances that your spot instances will be terminated.
         (default 1)
+  -spot_price_multiplier float
+        Multiplier for the spot price. Placing bids a fraction of a cent over the
+        current price would aggressively optimize for cost savings since the last
+        runtime hour before spot termination is free of charge.
+        (default 1.1)
```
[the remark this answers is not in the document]
cristim: The 1.0 will place spot instance bids at exactly the current market price, the 1.1 is there to offer a 10% buffer above the current price.

But I also find it a bit confusing, perhaps we should switch to percentages and also support absolute value increments, something like spot_price_buffer_percentage and spot_price_buffer_amount, what do you guys think?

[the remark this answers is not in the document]
cristim: @kartik894, @xlr-8 what do you think about my proposal to use percentage or absolut values for the spot price buffer?

I think something like spot_price_buffer_percentage=10(and make it a default when the aggressive policy is used) is more intuitive than the current spot_price_multiplier=1.1 for setting the bid to 10% over the current spot price. Also, negative percentage numbers don't make sense in this context, but the multiplier gives the users the chance to mess it up by giving 0.9, which will fail to launch any instances.

This is also quite consistent to the existing percentage/value concept which we use for keeping on-demand capacity, where we can have an absolute value as well as a percentage. This would bid X+current_price, where X may be given as a small fractional number, such as spot_price_buffer=0.001.

autospotting.go
```diff
@@ -124,6 +128,15 @@ func (c *cfgData) parseCommandLineFlags() {
 			"\tset your bid price to be higher than the on demand price to reduce the chances that your\n"+
 			"\tspot instances will be terminated.")
 
+	flag.Float64Var(&c.SpotPriceMultiplier, "spot_price_multiplier", 1.1,
+		"Multiplier for the spot price. Placing bids a fraction of a cent over the current price\n"+
+			"\twould aggressively optimize for cost savings since the last runtime hour before spot termination\n"+
+			"\tis free of charge.")
```
cristim: It turns out after the per-second billing the last hour is no longer free of charge. 

As of now the main benefit is that it protects the group from running spot instances that got significantly more expensive than when they were initially launched, but still somewhat less than the on-demand price.

For example:
- on-demand price: 1
- spot-price: 0.2 but increased to 0.9 over time
- this instance would be terminated by this new policy as soon as the price was exceeding 0.22 (using the vehiculated 10% price buffer we want to use as default the bid price would be 0.22, instead of 1).

With the default policy the spot instance in the above example may run forever after this significant price increase unless the price goes over 1.

This allows us to find other spot instance types that may be priced better as of the moment when the price increased.

core/autoscaling.go
```diff
@@ -536,6 +596,19 @@ func (a *autoScalingGroup) getAllowedInstanceTypes(baseInstance *instance) []str
 
 }
 
+func (a *autoScalingGroup) getPricetoBid(
+	baseOnDemandPrice float64, currentSpotPrice float64) float64 {
+	logger.Println("a.BiddingPolicy: ", a.BiddingPolicy)
+	if a.BiddingPolicy == "aggressive" {
+		logger.Println("Launching spot instance with a bid =", currentSpotPrice*a.region.conf.SpotPriceMultiplier)
+		return currentSpotPrice * a.region.conf.SpotPriceMultiplier
```
xlr-8: I realised that there might be a potential issue here, if the spot price is close to the on-demand price for some reasons (or the multiplier very high), then having the SpotPriceMultiplier might allow the bidding to go above that value. e.g.:

```bash
spotPrice: 0.5
onDemandPrice: 1
spotPriceMultiplier: 3.0
spotBid: => 1.5

# Other case
spotPrice: 0.9 (couldn't find better)
onDemandPrice: 1
spotPriceMultiplier: 1.5
spotBid: => 1.35
```

I think we might want to check that this potential maximum price isn't higher than the on-demand price, just for the sake of avoiding it; or we might end with spot instances running costing more than on-demand.

START.md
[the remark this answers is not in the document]
xlr-8: Your various comments made me think of: https://github.com/cristim/autospotting/pull/175/files#r156044213

We need to check the value anyway, just like it's done for the %/raw-value for on-demand instances. Also this value can't be set at an autoscaling-level if I'm not mistaken. Don't know if that's a goal, but if so some code need to be adapted

core/autoscaling.go
cristim: Good catch!

We can just take the minimum between ondemand price and the calculated spot price

autospotting.go
kartik894: I'm going ahead with percentage value as it seems more intuitive. Is that cool?

core/autoscaling.go
kartik894: Cool, will do.

terraform/autospotting.tf
```diff
@@ -4,6 +4,8 @@ module "autospotting" {
   autospotting_min_on_demand_number       = "${var.asg_min_on_demand_number}"
   autospotting_min_on_demand_percentage   = "${var.asg_min_on_demand_percentage}"
   autospotting_on_demand_price_multiplier = "${var.asg_on_demand_price_multiplier}"
+  autospotting_spot_price_buffer_percentage = "${var.asg_spot_price_buffer_percentage}"
+  autospotting_bidding_policy = "${var.asg_bidding_policy}"
```
xlr-8: Could run `terraform ftm -write=true` to format it properly

terraform/variables.tf
```diff
@@ -14,6 +14,16 @@ variable "asg_on_demand_price_multiplier" {
   default     = "1.0"
 }
 
+variable "asg_spot_price_buffer_percentage" {
+  description = "Percentage above the current spot price to place the bid"
+  default = "1.1"
```
xlr-8: Isn't it supposed to be `10.0`?

core/autoscaling.go
```diff
@@ -536,6 +629,30 @@ func (a *autoScalingGroup) getAllowedInstanceTypes(baseInstance *instance) []str
...
+	} else if a.region.conf.BiddingPolicy == "normal" {
+
+		logger.Println("Launching spot instance with a bid =", baseOnDemandPrice)
+		return baseOnDemandPrice
+
+	}
+
+	return baseOnDemandPrice
+}
+
```
xlr-8: I would also rewrite that as:
```go
func (a *autoScalingGroup) getPricetoBid(
	baseOnDemandPrice float64, currentSpotPrice float64) float64 {

	logger.Println("BiddingPolicy: ", a.region.conf.BiddingPolicy)

	if a.region.conf.BiddingPolicy == DefaultBiddingPolicy {
		logger.Println("Launching spot instance with a bid =", baseOnDemandPrice)
		return baseOnDemandPrice
	}

	logger.Println("Launching spot instance with a bid =", math.Min(baseOnDemandPrice, currentSpotPrice*(1.0+a.region.conf.SpotPriceBufferPercentage/100.0)))
	return math.Min(baseOnDemandPrice, currentSpotPrice*(1.0+a.region.conf.SpotPriceBufferPercentage/100.0))
}
```

Please note I removed the check of the value, as it seemed already done here: https://github.com/cristim/autospotting/pull/175/files#diff-de828bac595094756ea2a28fa415583bR219

My logic is always to avoid imbricated `if`, if there is an error check or direct return, then do it. I find that it usually simplifies reading/understanding of the code.

terraform/variables.tf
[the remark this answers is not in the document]
kartik894: Ah yes, my bad. Will change and ping you.

core/autoscaling.go
```diff
@@ -23,9 +22,24 @@ const (
...
 	// running in a group managed by autospotting.
 	DefaultMinOnDemandValue = 0
+
+	// DefaultSpotPriceBufferPercentage stores the default percentage value
+	// above the current spot price to place a bid
+	DefaultSpotPriceBufferPercentage = 10.0
+
+	// DefaultBiddingPolicy stores the default bidding policy for
+	// the spot bid on a per-group level
+	DefaultBiddingPolicy = "normal"
```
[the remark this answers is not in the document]
kartik894: Should I leave it as is in this pull request?

kartik894:
@xlr-8 @cristim Apart from running unit tests, I tested it on AWS too. Works fine.

core/autoscaling.go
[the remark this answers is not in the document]
xlr-8: I believe it's fine like this, I'll pass on those later on

xlr-8:
Cool, code looks good to me, please:
* fix the code climate issues
* rebase the PR for git log history purposes

On my side, I'll give it a quick shot tonight on AWS and I'll merge if everything worked as expected.

kartik894:
Will do the rebasing. The code climate points to cognitive complexity in two sections of the code which I'm not sure how to fix.

core/autoscaling.go
```diff
@@ -88,18 +117,78 @@ func (a *autoScalingGroup) loadConfOnDemand() bool {
...
+			if newValue > 0 {
+				a.region.conf.SpotPriceBufferPercentage = newValue
+				return done
+			}
+		}
+	} else {
+		debug.Println("Couldn't find tag", SpotPriceBufferPercentageTag)
+	}
+	return false
+}
```
xlr-8: I believe this could be rewritten as:

```go
func (a *autoScalingGroup) loadConfSpotPrice() bool {

	tagValue := a.getTagValue(SpotPriceBufferPercentageTag)
	if tagValue == nil {
		return false
	}

	newValue, done := a.loadSpotPriceBufferPercentage(tagValue)
	if !done {
		debug.Println("Couldn't find tag", SpotPriceBufferPercentageTag)
		return false
	}

	a.region.conf.SpotPriceBufferPercentage = newValue
	return done
}
```

Which would also solve code climate issue I believe, as the comment about imbricated if/else is something that adds complexity to the code and that code climate dislike.

core/autoscaling.go
```diff
@@ -88,18 +117,78 @@ func (a *autoScalingGroup) loadConfOnDemand() bool {
...
+					logger.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
+					return done
+				}
+			}
 		} else {
 			debug.Println("Couldn't find tag", tagKey)
 		}
 	}
 	return false
 }
```
xlr-8: Applying the same principle as before, it could be re-written as:
```go
func (a *autoScalingGroup) loadConfSpot() bool {
	tagList := []string{BiddingPolicyTag}
	loadDyn := map[string]func(*string) (string, bool){
		BiddingPolicyTag: a.loadBiddingPolicy,
	}

	for _, tagKey := range tagList {
		tagValue := a.getTagValue(tagKey)
		if tagValue == nil {
			debug.Println("Couldn't find tag", tagKey)
			continue
		}
		if _, ok := loadDyn[tagKey]; !ok {
			debug.Println("No method associated with tag", tagKey)
			continue
		}
		if newValue, done := loadDyn[tagKey](tagValue); done {
			a.region.conf.BiddingPolicy = newValue
			logger.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
			return done
		}
	}
	return false
}
```

Or considering we have only one policy, even as simple as:
```go
func (a *autoScalingGroup) loadConfSpot() bool {
	tagValue := a.getTagValue(BiddingPolicyTag)
	if tagValue == nil {
		debug.Println("Couldn't find tag", BiddingPolicyTag)
		return false
	}
	if newValue, done := a.loadBiddingPolicy(tagValue); done {
		a.region.conf.BiddingPolicy = newValue
		logger.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
		return done
	}
	return false
}
```

I presume both could help/should solve the codeClimate issue

kartik894: The first one still failed the CodeClimate run.... So, I used the second one.

xlr-8:
Tests on AWS worked as expected! Waiting for the rebase before giving it a final look/merge. Thanks again!

xlr-8:
All good! 👍 🎉

kartik894:
My pleasure 👍

Issue #176, opened 2017-12-03
Terraform deletion doesn't destroy role/policies


xlr-8:
## Issue type ##

- Bug Report

## Build number #

Independent

## Configuration ##

With latest version of terraform (>10.0), they split things by module instead of having all providers within the same binary.

```bash
> terraform --version
Terraform v0.10.6

Your version of Terraform is out of date! The latest version
is 0.11.1. You can update by downloading from www.terraform.io/downloads.html
```

## Environment ##

Independent

## Summary ##

When running the destruction of the lambda function, I hit the following issue:

```bash
Error applying plan:

1 error(s) occurred:

* module.autospotting.aws_iam_role.autospotting_role (destroy): 1 error(s) occurred:

* aws_iam_role.autospotting_role: DeleteConflict: Cannot delete entity, must detach all policies first.
	status code: 409, request id: 21abaff4-d837-11e7-a395-9be537ee44f3

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

## Steps to reproduce ##

Follow https://github.com/cristim/autospotting/blob/master/START.md#install-via-terraform instructions

## Expected results ##

Everything is deleted.

## Actual results ##

The role/policies aren't deleted due to this dependency in between them and the requirement to deattach the policies prior to role suppression.

PR #177, opened 2017-12-03
terraform: delete forces policies to detach

xlr-8:
When destroying the stack via terraform, it wouldn't destroy the
policies/roles, claiming the policies needed to be detached first.
With this option, the policies are detached and everything is correctly
deleted.

Fixes: https://github.com/cristim/autospotting/issues/176

   to it.
1. [ ] ~~Functionality not applicable to all users should be configurable.~~
1. [ ] ~~Configurations should be exposed through Lambda function environment
   and
   stacks defined as infrastructure code.~~
1. [ ] ~~Global configurations set from the infrastructure stack level should also
   support per-group overrides using tags.~~
1. [ ] ~~Tags names and expected values should be similar to the other existing
   configurations.~~
1. [ ] ~~Both global and tag-based configuration mechanisms should be tested and
  proven to work using log output from various test runs.~~
1. [ ] ~~The logs should be kept as clean as possible (use log levels as
  appropriate) and formatted consistently to the existing log output.~~


xlr-8:
I wanted to tick `No issues are reported when running make full-test` but I cannot as the master already has issues with that:

```bash
autospotting.go:22:1: comment on exported var Version should be of the form "Version ..."
autospotting.go:24:5: exported var ExpirationDate should have comment or be unexported
Found 2 lint suggestions; failing.
make: *** [Makefile:93: lint] Error 1
```

xlr-8:
Should be fixed via https://github.com/cristim/autospotting/pull/179

Issue #199, opened 2018-01-05
[Bug] The number of autoscaling group names that may be passed in is limited to 50.


sc-chad:
## Build: #
```
build nightly-515
```

## Report: #
```
Bug Report
```

## Region: #
```
us-east-1
```

## Summary #
```
Failed to describe AutoScaling groups in us-east-1 ValidationError: The number of group names that may be passed in is limited to 50.
```

## Cloudwatch Log shows: #

```
START RequestId: accf7de1-f264-11e7-bc7b-35986d9cfc05 Version: $LATEST
2018/01/05 22:06:26.895606 Starting autospotting agent, build  nightly-515 expiring on 06-Feb-2018
2018/01/05 22:06:26.895666 Parsed command line flags: regions='us-east-1' min_on_demand_number=0 min_on_demand_percentage=0.0 allowed_instance_types= on_demand_price_multiplier=1.00
2018/01/05 22:06:26 main.go:97: Scanning for available AWS regions
2018/01/05 22:06:27 main.go:69: Enabled to run in us-east-1, processing region.
2018/01/05 22:06:27 region.go:66: Creating connections to the required AWS services in us-east-1
2018/01/05 22:06:27 connections.go:29: Creating Service connections in us-east-1
2018/01/05 22:06:27 connections.go:43: Created service connections in us-east-1
2018/01/05 22:06:27 region.go:70: Scanning for enabled AutoScaling groups in  us-east-1
2018/01/05 22:06:27 region.go:280: Processing page 1 of DescribeTagsPages for us-east-1
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-1BA2W2WD0MGRY
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-30HA7GAP4WDU
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-1IZZ3P1EV6ENL
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-QTWL6STHTMI5
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-1CZIQRZ8R0RTQ
2018/01/05 22:06:27 region.go:282: us-east-1 has enabled ASG: AutoScalingGroup-1BHWNQV36U4NN
[...]

PR #201, opened 2018-01-10
Allow larger number of autoscaling group names

chaner:
# Issue Type

- Bugfix Pull Request

## Summary

https://github.com/cristim/autospotting/issues/199

   to it.
   and
   configurations.

### DIFF
--- a/core/region.go
+++ b/core/region.go
@@ -301,12 +301,9 @@ func (r *region) scanForEnabledAutoScalingGroups() {
 		return
 	}
 
-	svc := r.services.autoScaling
-
-	input := autoscaling.DescribeAutoScalingGroupsInput{}
 	pageNum := 0
-	err := svc.DescribeAutoScalingGroupsPages(
-		&input,
+	err := r.services.autoScaling.DescribeAutoScalingGroupsPages(
+		&autoscaling.DescribeAutoScalingGroupsInput{},
 		func(page *autoscaling.DescribeAutoScalingGroupsOutput, lastPage bool) bool {
 			pageNum++
 			logger.Println("Processing page", pageNum, "of DescribeAutoScalingGroupsPages for", r.name)
@@ -316,7 +313,7 @@ func (r *region) scanForEnabledAutoScalingGroups() {
 					name:   *asg.AutoScalingGroupName,
 					region: r,
 				}
-				if containsString(group.name, asgNames) {
+				if containsString(asgNames, group.name) {
 					r.enabledASGs = append(r.enabledASGs, group)
 				}
 			}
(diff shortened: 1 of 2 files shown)


core/region.go
```diff
@@ -303,9 +303,7 @@ func (r *region) scanForEnabledAutoScalingGroups() {
 
 	svc := r.services.autoScaling
 
-	input := autoscaling.DescribeAutoScalingGroupsInput{
-		AutoScalingGroupNames: asgNames,
-	}
+	input := autoscaling.DescribeAutoScalingGroupsInput{}
```
cristim: You can inline the `autoscaling.DescribeAutoScalingGroupsInput{}` below and delete this line entirely.

core/region.go
```diff
@@ -303,9 +303,7 @@ func (r *region) scanForEnabledAutoScalingGroups() {
 
 	svc := r.services.autoScaling
```
cristim: You can inline the `r.services.autoScaling` below instead of using this `svc` variable.

sc-chad:
Just wanted to comment.

Eric and I have been running this code in our environment since last night and has been working very well.  Memory usage has been consistent around 500mb with over 100 autoscaling groups.

sc-chad:
I had to increase the memory even before this change.  It was actually much higher prior.

cristim:
Can you guys do some profiling to see what uses all this memory?

Would setting the Lambda memory limit to 1GB be enough to mitigate this for your environment?

sc-chad:
That's what I currently have it set to.

xlr-8:
Wild guess, I can't remember where I read/commented that, but I feel like we might have duplicated data regarding, instances. Now I don't know how many, but depending on it - the memory usage might grow quite a lot. If someone remembers, please drop it here, cheers!

chaner:
@cristim @xlr-8 Thank you! I've made the suggested changes. Code climate still complains, but I think it's as simple as it's going to get without some refactoring (which I don't feel comfortable enough to do without digging into other parts of the codebase a lot more).

Thank you guys for helping us fix this issue!

xlr-8:
Despite the logic being simple, could you add 2 or 3 basic tests for the new method please? I'll glady merge it afterwards! Thanks again for the report & contributions, much appreciated!

xlr-8:
Bad news, your last commit cannot ensure tests are passing because the files aren't formatted with goimports/gofmt, some variables seem undefined, and some elements should be string pointers instead of strings.

But the good news is you can apply [this patch](https://gist.github.com/xlr-8/7779d80f04a1cca101b439a1039bd15a) an try again :)

chaner:
Aw, thank you! I was hoping to use travisci to run tests, but for some reason it was super slow. I'll apply your patch later today and try again.

chaner:
@xlr-8 thank you for your help. everything is green now!

Issue #174, opened 2017-12-02
Improve feature visibility of the project's features


xlr-8:
## Issue type ##

- Documentation Report

## Summary ##

Add an array/matrix of features / statuses.

## Details ##

When seeing various issues [tagged as Epic](https://github.com/cristim/autospotting/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3AEpic+) or even some such as: https://github.com/cristim/autospotting/issues/150, I feel that it could be good to have in our documentation an array or matrix of features being already done, not done, in progress etc.

That would not only increase the visibility of how rich the project is, but also provide quick insight on what has been done, or will be done/not done soon. Having it automatically generated based on the issue labeling/statuses would be great, as it could turn to be a pain to maintain.

cristim:
Here's a small wishlist I have from this:

- additional columns for having the feature exposed via Terraform and CloudFormation, since experimental features may not be exposed from the start, but just run as command-line options during development.
- a field for the maturity of each feature.

In the second table
- entries for  DefaultVPC, EC2Classic and VPC.
- some lines added about ECS, Kubernetes, Rancher, Mesos, Nomad, etc.(with links to docs and tools that allow draining the workload) such as https://github.com/kube-aws/kube-spot-termination-notice-handler, etc.

xlr-8:
OK, so I added couple elements you requested: https://github.com/cristim/autospotting/blob/hr-174/TECHNICAL_DETAILS.md#summary-of-supported-features--options

As mentioned in the commit message, I've also added "bugs" and the linked issue, to be more clear/transparent about it.

However, there isn't always doc referencing options & such, as mentioned in the commit.

Regarding the other desires, I'm a bit reserved.

**features exposed via Terraform/Cloudformation**
I don't think documenting features while developing is such a good idea, whether it is stable or it isn't. But I wouldn't document it as part of the "official documentation" - I'd rather have that in a changelog or something, then when stable/merged in the documentation.

**maturity**
How do you evaluate and maintain the maturity's level of a feature? By date of introduction? By number of bug reported about it? By feedback from the users?

**draining workload**
I've added Rancher, which made quite some sense to me; but I'm uncertain the link you provided is linked with autospotting - I haven't read it thoroughly, so I might have missed it - regarding the others, do you have specific links in mind?

cristim:
Looks great, thanks!

Regarding the others:

> **features exposed via Terraform/Cloudformation**
[...]

PR #200, opened 2018-01-10
adding documentation on how to get codedeploy to work with autospotting

sc-chad:
Wasn't sure where you wanted to link to the documentation so just adding a reference doc.

Closes: https://github.com/cristim/autospotting/issues/14

Issue #18, opened 2016-09-26
Working with Elastic Beanstalk


roeyazroel:
Just tested it with Elastic Beanstalk - working great. just add the tags during creation of new environment.

Maybe add to README?

Roey

cristim:
Thanks @roeyazroel for testing it, great to know it works!

It is actually mentioned in the Features, let me know if that's fine or should I make it more proeminent.

roeyazroel:
The ECS feature is go without saying, But the Elastic Beanstalk is more complected, has you can't update the ASG tags from the environment. You will need to find the ASG that the EBS is working with, and then manually update. every time the environment will be rebuild or solution stack will be updated, you might lose the tag. So updating the ASG tags from the environment configuration is safer. And the process of updating the environment tags isn't friendly at all.

cristim:
I understand, it would be great to have that Beanstalk stuff documented into a dedicated markdown page linked to the readme. Please create a pull request.

cristim:
Fixed in #34

cristim:
Beanstalk recently added the ability to easily tag resoiurces, see [here](https://aws.amazon.com/about-aws/whats-new/2017/10/aws-elastic-beanstalk-supports-ability-to-update-tags) 

I think we should update the documentation accordingly.

roeyazroel:
@cristim well, it's look like you can replace the entire section on beanstalk by "update the environment tags". 👍

cristim:
I think this can be closed

PR #203, opened 2018-01-12
doc: add table of supported feature/options tables

xlr-8:
# Issue Type

- Documentation Pull Request

## Summary

To gain a bit in visility about the evolution and/or support of existing
options and/or features 2 tables have been added to make it easier to
see.

Result looks like [this](https://github.com/cristim/autospotting/blob/hr-174/TECHNICAL_DETAILS.md#summary-of-supported-features--options).

Closes: https://github.com/cristim/autospotting/issues/174

   to it.
   and
   configurations.

### DIFF
--- a/TECHNICAL_DETAILS.md
+++ b/TECHNICAL_DETAILS.md
@@ -1,5 +1,54 @@
 # Technical Details #
 
+## Summary of supported features & options ##
+
+| Option  | Status | Override via ASG conf |
+| ------- | :----: | :-------------------: |
+| Run on multiple regions | :white_check_mark: | :heavy_minus_sign: |
+| [Keep a fixed minimum percentage of on-demand](https://github.com/cristim/autospotting/blob/master/START.md#minimum-on-demand-configuration) | :white_check_mark: | :white_check_mark: |
+| [Keep a fixed minimum number of on-demand](https://github.com/cristim/autospotting/blob/master/START.md#minimum-on-demand-configuration) | :white_check_mark: | :white_check_mark: |
+| Can bid at a certain percentage of the on-demand price | :white_check_mark: | :white_check_mark: |
+| Can bid the current spot price plus a certain percentage | :white_check_mark: | :white_check_mark: |
+| Whitelisting of certain instance types | [:beetle:](https://github.com/cristim/autospotting/pull/120#issuecomment-321856920) | :heavy_minus_sign: |
+| Blacklisting of certain instance types | :white_check_mark: | :white_check_mark: |
+
+For the options not directly linked to any specific part of the doc, please
+check the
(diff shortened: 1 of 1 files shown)


xlr-8:
My understanding is that the support for Beanstalk & CodeDeployed worked, but might need to be better documented as done here https://github.com/cristim/autospotting/pull/200 and reported there https://github.com/cristim/autospotting/issues/18

TECHNICAL_DETAILS.md
```diff
@@ -1,5 +1,29 @@
...
 
+## Summary of supported features & options ##
+
+| Option  | Status | Override via ASG conf |
+| ------- | :----: | :-------------------: |
+|Run on multiple regions | :white_check_mark: | :heavy_minus_sign: |
+|Keep a fixed minimum percentage of on-demand | :white_check_mark: | :white_check_mark: |
+|Keep a fixed minimum number of on-demand | :white_check_mark: | :white_check_mark: |
+|Can bid at a certain percentage of the on-demand price | :white_check_mark: | :white_check_mark: |
+|Can bid at a higher spot's price | :white_check_mark: | :white_check_mark: |
```
cristim: I'd rephrase this as something like "Can bid the current spot price plus a certain percentage"

TECHNICAL_DETAILS.md
```diff
@@ -1,5 +1,29 @@
 # Technical Details #
 
+## Summary of supported features & options ##
+
+| Option  | Status | Override via ASG conf |
```
cristim: How about some more columns for having the feature exposed via Terraform or CloudFormation, since experimental features may not  be exposed from the start, but just run as command-line options during development.

And also a field for the maturity of each feature

TECHNICAL_DETAILS.md
```diff
@@ -1,5 +1,29 @@
...
+|Can bid at a higher spot's price | :white_check_mark: | :white_check_mark: |
+|Whitelisting of certain instance types | :x: | :heavy_minus_sign: |
+|Blacklisting of certain instance types | :white_check_mark: | :x: |
+
+| Feature | Status |
+| ------- | :----: |
+|Install via Cloudformation | :white_check_mark: |
+|Install via Terraform | :white_check_mark: |
+|Works with code deploy | :white_check_mark: |
+|Works with beanstalk | :white_check_mark: |
```
cristim: I'd like to have some lines added about ECS, Kubernetes, Rancher, Mesos, Nomad, etc.(with links to docs and tools that allow draining the workload)
- Also DefaultVPC, EC2Classic and VPC.

xlr-8:
The branch needs to be updated following the merge of [this PR](https://github.com/cristim/autospotting/pull/204/files). I'll take care of it

PR #205, opened 2018-01-12
 iam: add 'CreateServiceLinkedRole' to permissions 

xlr-8:
# Issue Type

- Bugfix Pull Request

## Summary

This permission seems required in order to have autospotting running fine
with CodeDeploy and ECS in order to avoid facing AuthFailure errors.

In order to eventually make it easier to work with terraform, the
autospotting_role's name has been exported.

Closes: https://github.com/cristim/autospotting/issues/187

   to it.
   and


cristim:
I couldn't reproduce this issue, is this still a problem for you guys?

twoism-dev:
Hi @cristim 

Sorry for the late response - this fix works for myself and @homme at FATMAP

Would it be possible to merge?

Ben

cristim:
@twoism-dev, @homme: I know the fix works.

My point was that it may have already been fixed by AWS in the meantime so fixing it on our side is redundant.

I'd like to see if there's anyone else out there having this issue on a fresh installation.

twoism-dev:
@cristim 

Yes - I've just provisioned a new environment with the latest version of autospotting - and had the same issue, which was fixed by this commit.

cristim:
Fair enough, thanks for reporting it.

Issue #170, opened 2017-12-02
Switch to basic Go environment for Lambda


xlr-8:
## Issue type ##

- Feature Idea

## Summary ##

Switch the project to the default Lambda go runtime provided by AWS.

## Details ##

Since the AWS re:invent, the Go support for Lambda has been announced, so instead of using tricks to build a binary and then make it run under Python runtime, it would be best to use directly the runtime environment provided by AWS.

Note: it is not yet available in eu-west-1 console for example, so need to wait.

cristim:
As far as I can see the native go support for Lambda is not yet generally available. 

But we should definitely switch to it once it is released.

cristim:
I had a look at the official Go Lambda runtime and it seems like it may take a bit of work to get it also build as a command line tool, but the Lambda function build process looks significantly simpler than what we have now.

xlr-8:
I have made an updated version that seem to run fine so far with Terraform - Cloudformation was also updated with the same logic but not yet tested.

I have to double check the Makefile rules and verify I can run it locally - or find a way to do so - and we should be good.

cristim:
Worked like a charm, huge thanks for this awesome contribution!

xlr-8:
Soon X-Ray coming! :smiley:

PR #207, opened 2018-01-21
lambda: make it runs in go1.x rather than python2.7

xlr-8:
# Issue Type

- Enhancement Pull Request

## Summary

The code was currently running as a binary called via python2.7,
however since the annoucement of go support for lambda by AWS, it makes
more sense to conf/push it directly into that environment and benefit
from both the performance, support and monitoring/tracing.

Closes: https://github.com/cristim/autospotting/issues/170

   to it.
   and
   configurations.


autospotting.go
```diff
@@ -26,7 +26,7 @@ var Version string
 var ExpirationDate string
 
 func main() {
-	run()
+	lambda.Start(Handler)
```
cristim: I think we can use one of the [reserved environment variables](
https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html) to determine if the binary is executed in the Lambda environment so we then run the handler, and keep calling the `run()` when executed locally, like we do now. A simple if like this one should do the trick (note: this is just a hunch, and it wasn't yet tested): 

```
if os.Getenv("AWS_LAMBDA_FUNCTION_NAME") != "" {
  lambda.Start(Handler)
} else {
 run()
}
```

xlr-8: Exactly my thought

xlr-8:
The small fix for checking environment seem to work just fine, I haven't tested cloudformation yet though.
I'd like someone to double check it as well if possible.

cristim:
I tested it with CloudFormation and it worked like a charm once i updated my stack configuration.

PR #210, opened 2018-01-31
remove 25% cap on allowedinstancetypes

JaveriaK:
Signed-off-by: Javeria Khan <jkhan@branch.io>

# Issue Type

- Bugfix Pull Request

## Summary
When choosing to run autospotting with AllowedInstanceTypes; the goal is to stick to a few specific types (even one); and have all of the on-demands in an ASG converted to those types. 

At the moment the code caps the number of spots in an ASG to stay below 25% for any particular type. This changeset skips that capping check when AllowedInstanceTypes has been defined.

I tested this in a custom build running in CloudFormation. Where previously, using AllowedInstanceTypes would stop converting On-Demands after a few instances (25% according to code); this changeset now allows all of the on-demands to be converted. 

   to it.
   and
   configurations.


JaveriaK:
Yes `allowedInstanceTypes` has been working fine for me. I've used it across a few different type of asgs. The only thing is it sticks to creating the same type of instance as you already have in the asg, and since there will usually be only one type, it will create only that one type. Which is fine for my use case.

I agree, the logic for this feature should be updated to disallow empty spaces & nil. Probably not in the same PR though, as that should be a separate commit.

cristim:
This can also be an issue when the algorithm only has a single available instance type due to market conditions, other people reported it in the past. 

A proposed solution was to randomly pick the on demand instance we want to replace, otherwise we can get stuck replacing the same one that is already over the 25% limit.

But restricting the logic to a small number of instance types makes it much more likely to happen.

JaveriaK:
That's reasonable. The logic should probably be updated to circle in between the allowed instance types without sticking to one particular one, which is what it does now.

cristim:
@JaveriaK That's pretty much what we do so far:

- we iterate over all the EC2 instances in the group and we pick the first on-demand one we find in the list of members
- then we try to find a suitable spot instance type, iterating over all possible types that are compatible with this instance.

The problem is that the first on-demand instance returned by the API is always the same, so if we can't find any suitable instance types for replacing it, for whatever reasons, we just stop replacing instances for that group and we get stuck.

Those reasons can be multiple:
- prices are higher than the current price for all the allowed and compatible instance types
- or we already have in the group enough instances from that type (this is the 25% cap currently implemented for safety reasons), so getting more of them can increase the odds of impact in case that spot instance type is oubid in that AZ and we'd lose significant capacity.

Doing the math, this point will be hit quite early on regions with just 2 AZs, where we get stuck like this after replacing just 25% of the nodes.

It makes a lot of sense to randomize the instance we try to replace, to eventually get those 25% instances from all the AZs.

In the regions with 4 AZs(I think us-east-1 may be the only one) we'd be able to replace all of them, but in regions with just 2 AZs we'd still stop at 50% of the capacity even with random instance selection. I'm inclined to think that this is desirable for those regions for safety reasons.

As a user I'd like to be given the safest, even if not the most cost-effective option by default, but to be given the chance to opt-in if I'm feeling adventurous.

So in my opinion a proper implementation for addressing this feature would look like this:
- implement randomized selection when replacing on-demand instances to avoid getting stuck too early because of this cap 
- keep the 25% cap on by default, but allow people to disable it by global or group-based configuration if they are feeling lucky
- document this behavior in the docs/FAQ and recommend users to keep some on-demand capacity using the other configuration options we have for this, to reduce the risk of total outage in those regions.

xlr-8:
Your last phrase was exactly where I was going to:

> On another hand, we do have the option to configure exactly how much on-demand capacity to keep, so we could just rely on that setting and avoid setting any cap on a per instance type basis.

So the question here is:
`What would be the consequence of removing the capping limitation in favor of only managing it via the "on-demand" parameters. Considering the instance prices are not always the same per AZ and that the variation can be on only 1 AZ out of X`.

Based on this, I believe we could remove such limitation as it wouldn't represent too much risk.

The modifications of the previously mentioned behavior (picking up random instance rather than first) would still provide an improved logic of replacement, as when a price is too high in 1 AZ, we might get stuck and stop trying to replace instances no matter how big/small the ASG capacity is - which is not desired.

cristim:
My subconscious has been running this in a background thread over the last few days and I think we can implement resilience similar to what was intended by this 25% cap by randomly selecting the spot instance type from more compatible candidates if they are priced close enough to the minimum price.

Let's say the hourly on-demand is $1, and we have multiple candidates priced around 0.2, with this in mind we could pick any of them, even if it costs 0.21 or so while the minimum we would try to enforce with the 25% cap was 0.195.

This won't make any difference in the current cases in which we get stuck, but it would be safer than the 25% cap implementation when we have multiple candidates priced within a small range.

cristim:
@JaveriaK let's get rid of the cap entirely.

JaveriaK:
Sorry for the late reply, so I just read through the conversation. To confirm, we want to remove the capping logic entirely and let it select whatever instances it finds in the allowed price range?
I agree with your analysis, sticking to similar compatible instances should generally be a better idea for most applications.

JaveriaK:
Would you like me to update instance_test.go as well and remove the logic that uses isSpotQuantityCompatible?

cristim:
Yes please, and the tests should be green

JaveriaK:
Yep, was trying to be as less disruptive as possible. Looks okay now.

Issue #213, opened 2018-02-08
c5 with ebs optimization should not be replaced by c3


russellballestrini:
# Github issue #

```
2018/02/08 02:29:23 instance.go:273: Comparing c3.2xlarge with c5.2xlarge
2018/02/08 02:29:23 instance.go:116: Comparing price spot/instance:
2018/02/08 02:29:23 instance.go:120: EBS Surcharge : 0.05
2018/02/08 02:29:23 instance.go:123: Spot price: 0.47
2018/02/08 02:29:23 instance.go:124: Instance price: 0.34
```

## Issue type ##

- Bug Report

## Build number ##

```text
latest nightly
```

## Configuration ##

I turned on debugging

## Environment ##

- us-east-1
- VPC

## Summary ##

`c5.2xlarge` is being swapped out for `c3.2xlarge` even though it is more expensive, when EBS Optimization is enabled.

cristim:
We eventually traced this down to differences in the spot price data, depending on which spot product we use to request the prices. For relatively recent instance types that don't support VPC Classic, we get no price data when using **Linux/Unix**, but only with **Linux/Unix (Amazon VPC)**.

@russellballestrini is now preparing a patch to make the spot product configurable and we'll have a default value that supports Linux on VPC out of the box.

In order to support VPC, EC2 Classic and even Windows in the same account, a possible workaround would be to install AutoSpotting multiple times, and couple this feature with the one introduced in #209 (once both these get merged) and use dedicated tags to make the distinction between Classic, VPC and Windows stacks for each of these AutoSpotting parallel installations configured each with a different spot product.

russellballestrini:
resolved by #214

PR #214, opened 2018-02-08
Make spot product description config

russellballestrini:
# Issue Type 

- Feature Pull Request
- Documentation Pull Request

## Summary

The `Linux/UNIX` product description does not have the price history for the most recent 2 instance type generations, as proved here: http://pad.yohdah.com/598/spot-product-no-history

As a result we have decided to change the default SpotProductDescription to `Linux/UNIX (Amazon VPC)` as well as make it configurable for backwards compatibility reasons (and so that we can support other operating systems like Windows, etc.)

* Changes the default `Spot Product Description` from `Linux/UNIX` to `Linux/UNIX (Amazon VPC)`
* Adds a flag to allow configuration of `-spot_product_description`
* Updates CLI help text for readability
* Derives flag defaults from constants 

   to it.

 Might fix the following: #206 #213 

  This doesn't work on Mac but I'm pretty but I ran `go fmt` and `go test`
 
   and
   configurations.


xlr-8:
I would have simply added a small check as we have the "valid values" in order to ensure it's included, or fallback on the default one. I can open an issue if you think this is relevant

PR #218, opened 2018-02-14
Fix Terraform typo

gozer:
Missing closing '"'


cristim:
@gozer  good catch and thanks for your contribution to the project!

PR #221, opened 2018-02-21
Support glob matching on allowed instance type

atillamas:
# Issue Type

- Bugfix Pull Request

## Summary

According to documentation allowedinstancestype should support glob, this is not the case but this PR fixes this


atillamas:
@cristim added, hopefully you could take a quick look

cristim:
@atillamas I agree with your proposal about line 219, let's make this logic consistent between the allowed and disallowed cases.

atillamas:
@cristim Updated the logic

xlr-8:
I'm not against it, but as we are to update the checking for `allowedInstanceTypes`, I'd also do the same modifications (loading/checking) as for `diasallowedInstanceType`, such as [this](https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L623-L667).

That would allow:
1. To be able to give ASG configuration the priority, which is our main logic
1. Remove empty elements in the array

A couple test might need to be updated as the logic differs, but shouldn't be anything too big, would you be fine with doing such modifications @atillamas ? If you need I can provide further help

atillamas:
@xlr-8 I'm not really sure i follow, all that is already in place? All the fix does is matching the glob pattern with every instance to see if it matches,  previously it would just match the full name ie m3.xlarge but not m3.* as stated in the documentation (Spent hours trying to figure out why it didn't launch any instances), i don't see what else is needed to be added to be honest.

xlr-8:
@atillamas, your PR is fine.

My point was that all of our options loading logic is: take CLI parameters, or a default value if not set, and if this is set in the ASG overide the configuration, see https://github.com/cristim/autospotting/blob/master/TECHNICAL_DETAILS.md#summary-of-supported-features--options

The logic for `allowedInstanceType` differs by the fact that the CLI has the priority, which isn't the case for the `disallowedInstanceType` I would have liked to have that uniformed, as we were to update some parts of this option here, and it is a 5-10min modification, I assumed it would be possible for you to do it. But if you prefer not to, fine with me.

Eventual update could be to add the globbing possibility in the descriptions:
https://github.com/cristim/autospotting/blob/master/terraform/variables.tf#L2
https://github.com/cristim/autospotting/blob/master/cloudformation/stacks/AutoSpotting/template.json#L84

atillamas:
@xlr-8 ok thanks for clarifying, but it didn't work when i tried the allowedInstanceType from the CLI either, but according to https://github.com/cristim/autospotting/blob/master/START.md 

> $ ./autospotting -h
Usage of ./autospotting:
  -allowed_instance_types="":
        If specified, the spot instances will be of these types.
        If missing, the type is autodetected frome each ASG based on it's Launch Configuration.
        Accepts a list of comma or whitespace seperated instance types (supports globs).
        Example: ./autospotting -allowed_instance_types 'c5.*,c4.xlarge'

And thats why i assumed it was just a bug and not new feature that was needed to be added/changed.
I'll see if i can figure out what needs to be changed, but I won't have much time this weekend. Maybe during next week.

xlr-8:
@atillamas thanks! Again if you don't have time it's fine, we could merge this PR.
I believe you could check the `disallowTypes` way of doing to copy it, that should be just fine :+1:

atillamas:
@xlr-8 I hope i got it right now, had to leave if allowed == "current" in there otherwise a test would fail. Otherwise it's more or less a copy of the disallowedTypes.

xlr-8:
Great! Thank you a lot!
I was a bit curious that the all tests passed with such configuration, while it shouldn't in my head; so I checked them out, and fixed/added some of those via: https://gist.github.com/xlr-8/7779d80f04a1cca101b439a1039bd15a - one tag value was outdated and couple cases were added to match `disallowedInstanceTypes`

It would be good to apply it I believe, as it would ensure the good functioning/strength of our loading logic.

atillamas:
@xlr-8 Now it's applied. Thank you!

PR #222, opened 2018-02-22
Spot Instance Holding Request cancelling.

tootedom:
# Issue Type

- Feature Pull Request

## Summary

Briefly discussed this on the group.  Let me know what you think.

When a spot request is issue, the normal operation is for this request to go into pending
by AWS until it is fulfilled.  However, there are occasions when the spot request is put
into a 'holding' state by amazon.  An example would be, there is no capacity available for
instance type requested.  A list of the type of 'holding' status can be found on the following
page: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-bid-status.html

By default when a spot request is in holding, it will remain there until it is fulfilled.
This can be hours.  

I've added the parameter: `-max_time_spot_request_can_be_holding`
that can be set, to the number of seconds (since the spot request was created), that the
spot request can be in this state.  If it exceeds these number of seconds the spot request
is cancelled.

When the spot instance is in the holding state, the request isn't waited on, and instead we wait for the next run of the lambda to check the spot request.

Also adding in checking for err on the `waitForAndTagSpotInstance()`, so that activeSpotInstanceRequest isn't set and `activeSpotInstanceRequest.InstanceId` is nil when we attempt to reference it.

also noticed that CancelSpotInstanceRequest wasn't in the list of permissions, so I've added that.

I'm sure I've missed some things.

   to it.
   and
   configurations.


tootedom:
I'll check out the checkout the coveralls issues, and the coverage, and will update the pull request later on.

tootedom:
@cristim yup, intend to look at all the reports.  I have ideas on how to fix a few of the issues and
Will update in a few days.  

Cheers
/Dom

tootedom:
hola,

Apologies for the delay, in pushing the changes.  Let me know what you think.

cheers
/dom

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -160,6 +166,7 @@
                 "ec2:DescribeSpotInstanceRequests",
                 "ec2:DescribeSpotPriceHistory",
                 "ec2:RequestSpotInstances",
+                "ec2:CancelSpotInstanceRequests",
```
cristim: Have you considered combining the `ValidUntil` parameter of [ec2.RequestSpotInstancesInput](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RequestSpotInstancesInput) instead of handling all this ourselves?

This should considerably simplify the below logic, the spot instances would be launched and their request would be cleaned up automatically by AWS after it expires. Any launched instances would be kept running after the spot instance requests have been cleaned up.

cloudformation/stacks/AutoSpotting/template.json
```diff
@@ -93,6 +93,11 @@
       "Default": "",
       "Description": "Comma separated list of tag=value on which to filter the ASGs that autospotting considers.  By default (if no filters are specific) then 'spot-enabled=true' is used.  Example: 'spot-enabled=true,environment=dev'",
       "Type": "String"
+    },
+    "MaxTimeSpotRequestCanBeHolding": {
+      "Default": "0",
+      "Description": "The amount of time (in seconds) that a spot request can remain open, but in a holding state by amazon, before it is cancelled.  The default is 0, which means the request will not be cancelled",
```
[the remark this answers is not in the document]
tootedom: will update

cloudformation/stacks/AutoSpotting/template.json
tootedom: Hi there,

The ValidUntil, is a good point, but it won't address the other issue.  We need to look at the status of the spot instance request to see if it is in the holding state.  

Currently we wait on the spot instance request regardless, and this times out the lambda, and produces an invocation error (timed out).  From experience, I've seen requests in holding be there for hours.  This results in "errors" for hours.  Errors should really be an exceptional case, not a normal; as an error means it should be alerting someone to look into the issue.

Therefore for me, we still need to check the status of the request and not wait on it (just check it on the next lambda run).  

The ec2:CancelSpotInstanceRequests was already required, and missed, the following was pre-existing:

- https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L595

autospotting.go
```diff
@@ -30,31 +30,26 @@ func main() {
...
+		"allowed_instance_types=%v disallowed_instance_types=%v "+
+		"on_demand_price_multiplier=%.2f spot_price_buffer_percentage=%.3f "+
+		"bidding_policy=%s tag_filters=%s "+
+		"spot_product_description=%v max_time_spot_request_can_be_holding=%d",
+		conf.Regions,
+		conf.MinOnDemandNumber, conf.MinOnDemandPercentage,
+		conf.AllowedInstanceTypes, conf.DisallowedInstanceTypes,
+		conf.OnDemandPriceMultiplier, conf.SpotPriceBufferPercentage,
+		conf.BiddingPolicy, conf.FilterByTags,
+		conf.SpotProductDescription, conf.MaxTimeSpotRequestCanBeHolding)
```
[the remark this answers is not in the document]
tootedom: I will do.  Although, we will eventually hit the 25 line limit on the codeclimate.

cloudformation/stacks/AutoSpotting/template.json
cristim: The `validUntil` would be set to the maximum amount of time you want to allow instances in the holding state, for example five minutes.

If we didn't manage to launch any instances before this expires, the request is immediately cancelled by AWS without any action required on our side.

If the instance was launched, it would be kept running but the spot request would be cancelled nevertheless. I think this also has some potential to improve the throttling situation which happens quite often on large installations (50+ groups handled in parallel).

But we need to pay attention not to get any orphaned instances outside the groups, if their spot requests have since been cancelled. One way to do this is to also tag the instance with the group it needs to be added to, instead of just tagging the SIR

tootedom: but the ValidUntil is going to effect all spot instance requests that are made.  You are going to end up with a lot of spot instances, and cancelled spot requests.  Which is going to confuse users.  i.e. if I have a running spot instances then I expect an active spot instance request.

cristim: That's not the case even today, the SIRs are cancelled after a week (the default value for `ValidUntil`) even if their instances keep running for longer time.

So as far as I can tell the SIR is meant to be temporary, only useful until an instance can be launched from it, then it can be safely discarded.

The source of truth should always be the running instances.

tootedom: Ah ok, we are using it in our Dev environments that shutdown daily, so nothing lasts a week.

So.  Given this, what do we wish to do:

- when a spot request is seen to be in holding set the valid until (can you set this on a existing request - or does it have to be done when the request is made?)

- can we tag the instance with the spot request it was created from?

The tagging of instance for which no spot request exists (is cancelled due to the valid until), sounds like quite a bit of work

tootedom: I say this is a lot of work, as the code base (core/autoscaling.go) is all about iterating open or active SIRs

tootedom: I don't see the ability to update a spot request after creation (only fleets, not the same thing).

I don't see valid-until working without a lot of work.

tootedom: Ok.  Do we wish to change:

-max_time_spot_request_can_be_holding=0:

To:

-seconds_spot_request_valid_for

And then change the spot request creation to use this to create the specified valid-until on the spot request?

----
I'm very unsure about the following:

```But we need to pay attention not to get any orphaned instances outside the groups, if their spot requests have since been cancelled. One way to do this is to tag the instance with the group it needs to be added to, instead of tagging the SIR. The instances are described already across the entire fleet, so we don't need to do any additional API calls, just to verify if they are tagged for the current group.

Unfortunately we need to wait for the SIR to launch instances because we need to make sure they have been tagged, because the SIR won't tag the instances for us and we have to do it as soon as they were launched.```

Comment (cristim) [reply cloudformation/stacks/AutoSpotting/template.json]: I think the logic would have to enforce that SIRs are temporary and never live longer than a Lambda function run.

- We just launch the SIR and wait for the instance to come up using the waiter we already have in place (only so we can tag the instance), and clean up the SIR as soon as their instance was launched. We can also set the SIRs to be cancelled automatically after less than the 5 minutes Lambda timeout, just in case we have a crash so we don't have SIRs lying around.
- We remove all the logic of tracking and handling open SIRs between multiple function runs, as well as the tracking of intermediate SIR states(handled by the waiter anyway within the run) and also remove the cancelling logic you just introduced. This should simplify things a lot.
- I think it makes little sense to have the `-seconds_spot_request_valid_for` configurable, except for testing purposes. Most of the time a ~4min default should be fine.
- We tag the instance with the group it needs to be added to, not with the SIR. The next run will see the spot instances that have to be added to the current group and do the swapping.

This will be some work, but mainly cleaning up and simplifying the logic, I expect it to remove a few hundreds lines of code which is always great.

Comment (tootedom) [reply cloudformation/stacks/AutoSpotting/template.json]: ```We just launch the SIR and wait for the instance to come up using the waiter we already have in place (only so we can tag the instance)```

Waiting for instances on an open, but in holding by Amazon, will cause the lambda to hang and time out (i.e. an Error).   We would still have to check for "open" but in holding, so we don't unnecessarily wait on SIR in holding

----

```I think it makes little sense to have the -seconds_spot_request_valid_for configurable, except for testing purposes. Most of the time a ~4min default should be fine.```

If we set the default for 4 mins, then the logic for finding SIR and instances all has to change (`havingReadyToAttachSpotInstance()` in autoscaling.go and logic around it); and I would say that having it configurable would turn out to be something required.

----

All of the above is really a different requirement:

- don't use SIR as the means for adding instances to ASGs

Yup, it can (if we are careful) address the don't waiting on stuff that Amazon puts in holding.  

This pull is specifically about improving the current situation.  It waits unnecessarily on SIR's in holding.  The default operation is as it was previously (never cancel a SIR that is in holding).  The change is to improve it by not waiting on the held SIR.  

Given this. 

If I was to update the pull request to not include the command line option for

- max_time_spot_request_can_be_holding

and update it so that we don't cancel the spot and don't compare time it has been open. So that nothing leaks into the public interface of the lambda (all behind the scenes changes)

Then we get:

- missing ec2:CancelSpotInstanceRequests fix
- nil pointer exception fix by checking for errors on the waitForAndTagSpotInstance(), and not derefencing a nil id: (https://github.com/cristim/autospotting/blob/master/core/autoscaling.go#L584)
- not waiting on holding requests.

Then I think as a separate pull request I can potentially look into the more complex task/requirement of:

- don't use SIR as the means for adding instances to ASGs


Does that make sense?

Comment (cristim) [reply cloudformation/stacks/AutoSpotting/template.json]: Sure, let's do this!

@xlr-8 what are your thoughts on this?

Comment (tootedom) [reply cloudformation/stacks/AutoSpotting/template.json]: Just pondering if I'm good to do the baby steps?  I might get a couple of free hours tmr.

Comment (cristim) [reply cloudformation/stacks/AutoSpotting/template.json]: By all means!

Feel free to get in touch on gitter if you have any questions or concerns.

Huge thanks for volunteering to do this, it'll be a great improvement.

**Comment by tootedom** [comment]
Hi there.  

Again, apologies for the delay (holidays.. get in the way of progress on things :-).  I've update the pull to remove exposing the max time for a spot request can be open, and have done the minimal changes to address:

- not waiting on open and held requests
- handle aws failures on the waiting for a spot request instance
-- caters for the nil pointer issue.

Minimum to improve the current issues.
----

For the other items we've discussed, I'll work on a completely separate pull for:

- don't use SIR as the means for adding instances to ASGs

Unless you have made progress on this already.  I have a 4 hour bus ride tmr, and a plane flight (in coming future); so I can have a look into this.

cheers
/dom

File: autospotting.go [hunk 4]
```diff
@@ -30,20 +30,15 @@ func main() {
...
 }
 
-func run() {
-
-	log.Println("Starting autospotting agent, build", Version)
-
-	log.Printf("Parsed command line flags: "+
-		"regions='%s' "+
+func outputParsedCommandLineOptions() {
+	log.Printf("Parsed command line flags: regions='%s' "+
```
cristim: Why change this formatting?

tootedom:
Fixed the formatting (apologies).
I don't see any merge conflicts to resolve

cristim:
Not sure what the problem is, but I can't merge this:

image

tootedom:
I'll see if I can fetch master again.  Maybe my 4g fell off

tootedom:
I see this:
screen shot 2018-03-02 at 17 46 56

Maybe I need to recreate the branch from master....

tootedom:
leave it with me.  I'll create a brand new branch from master, and apply the changes on it.  and send an new pull request.  I'll close this one when I've done that.  Not gonna be till tmr though.  Sorry about this.

/dom

tootedom:
see new pull #228  I'll close this.
thanks
/dom

cristim:
Thanks for taking care of this, I'll review #228 shortly.

PR #224, opened 2018-02-26
Terraform s3 support

### LINKED ISSUE #223: Support installing Lambda code from S3
# Github issue #

## Issue type ##

- Feature Idea

## Build number #

```text
965aa93
```

## Summary ##

We have a CI pipeline and build the Lambda code at the start, push to S3 and then install into different environments by pulling down from S3, in the same way that you might use a more traditional build artefact (RPM, DEB, JAR, etc. etc. etc.)

It would be nice to be able to use the Terraform code provided here but to pass the `s3_bucket` and `s3_key` parameters to the `aws_lambda_function` resource, instead of referencing a local file.

Is this something that you'd be keen/happy to see as a feature? If so, I'll happily raise a pull request.

grahamlyons:
# Issue Type

- Feature Pull Request

## Summary

Fixes #223 

   to it.
   and
   configurations.


Makefile
```diff
@@ -103,6 +105,18 @@ travisci-checks: fmt-check vet-check lint                    ## Pass fmt / vet &
 travisci: archive travisci-checks travisci-cover             ## Executed by TravisCI
 .PHONY: travisci
 
+terraform-test: archive .bin/terraform                            ## Test the Terraform code
+	@.bin/terraform init terraform
+	AWS_DEFAULT_REGION=us-east-1 .bin/terraform validate -var lambda_zipname=$(LOCAL_PATH)/lambda.zip terraform/
+	AWS_DEFAULT_REGION=us-east-1 .bin/terraform validate -var lambda_s3_bucket=bucket -var lambda_s3_key=key -var lambda_zipname= terraform/
```
cristim: There's a whitespace after the equals sign in `lambda_zipname= terraform/`

xlr-8:
Hey @grahamlyons, thanks for you work! Please label the issue with `Need a review` whenever you think you're done! I'll gladly take a look!

grahamlyons:
Thanks @xlr-8, I'm finished now but I don't think I have label permissions on here.

xlr-8:
Cool, I'll check it out this week!

Makefile
```diff
@@ -8,6 +8,7 @@ BUCKET_NAME ?= cloudprowess
 FLAVOR ?= custom
 LOCAL_PATH := build/s3/$(FLAVOR)
 LICENSE_FILES := LICENSE
+TERRAFORM_VERSION := 0.11.3
```
xlr-8: Although it's good to stick to one version, I don't know how to feel about all the upgrades of terraform, as they upgrade it at a fairly good rhythm.

Makefile
```diff
@@ -103,6 +105,18 @@ travisci-checks: fmt-check vet-check lint                    ## Pass fmt / vet &
 travisci: archive travisci-checks travisci-cover             ## Executed by TravisCI
 .PHONY: travisci
 
+terraform-test: archive .bin/terraform                            ## Test the Terraform code
+	@.bin/terraform init terraform
+	AWS_DEFAULT_REGION=us-east-1 .bin/terraform validate -var lambda_zipname=$(LOCAL_PATH)/lambda.zip terraform/
+	AWS_DEFAULT_REGION=us-east-1 .bin/terraform validate -var lambda_s3_bucket=bucket -var lambda_s3_key=key terraform/
+.PHONY: terraform-test
```
xlr-8: The indent of the comment `## Test the Terraform code` would have to be reduced, you can type `make help` to understand why :)

terraform/autospotting/lambda.tf
```diff
@@ -1,27 +1,27 @@
...
+  autospotting_disallowed_instance_types = "${var.autospotting_disallowed_instance_types}"
+  autospotting_min_on_demand_number = "${var.autospotting_min_on_demand_number}"
+  autospotting_min_on_demand_percentage = "${var.autospotting_min_on_demand_percentage}"
+  autospotting_on_demand_price_multiplier = "${var.autospotting_on_demand_price_multiplier}"
+  autospotting_spot_price_buffer_percentage = "${var.autospotting_spot_price_buffer_percentage}"
+  autospotting_spot_product_description = "${var.autospotting_spot_product_description}"
+  autospotting_bidding_policy = "${var.autospotting_bidding_policy}"
+  autospotting_regions_enabled = "${var.autospotting_regions_enabled}"
+  autospotting_tag_filters = "${var.autospotting_tag_filters}"
 }
```
xlr-8: Could you `terraform fmt -write=true terraform/autospotting/lambda.tf` please? Just for the sake of having it already formatted

Makefile
cristim: I guess this is just because we need to create the download URL.

I would prefer to run it from a container, using the latest available Terraform Docker image.

Makefile
```diff
@@ -103,6 +105,18 @@ travisci-checks: fmt-check vet-check lint                    ## Pass fmt / vet &
...
+
+.bin/terraform:
```
cristim: Why not use a Docker container?

grahamlyons:
> However why making it a module?
@xlr-8, the reason this is done is because the `filename` and `s3_*` parameters on the `aws_lambda_function` Terraform resource conflict with one another. The separate module takes care of detecting which parameters have been passed and provisions the Lambda function one way or the other and then passes back the ARN etc. of whichever one was used.

.gitignore
```diff
@@ -9,3 +9,5 @@ handler.zip
 terraform/.terraform/
 terraform.tfstate*
 Makefile.lambda
+.bin/
```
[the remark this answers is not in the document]
grahamlyons: Good catch. Removing it.

Makefile
```diff
@@ -24,6 +25,7 @@ all: fmt-check vet-check build test                          ## Build the code
 clean:                                                       ## Remove installed packages/temporary files
 	go clean ./...
 	rm -rf $(BINDATA_DIR) $(LOCAL_PATH)
+	rm -rf .bin .terraform
```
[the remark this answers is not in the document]
grahamlyons: Ta. Will ditch it.

Makefile
```diff
@@ -8,6 +8,7 @@ BUCKET_NAME ?= cloudprowess
 FLAVOR ?= custom
 LOCAL_PATH := build/s3/$(FLAVOR)
 LICENSE_FILES := LICENSE
+TERRAFORM := docker container run --rm -e AWS_DEFAULT_REGION=us-east-1 -v $(shell pwd):$(shell pwd) -w $(shell pwd) hashicorp/terraform
```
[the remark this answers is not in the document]
grahamlyons: This is the new style of command, introduced in Docker 1.13. `docker run` still works but it's now grouped under `docker container ...` which is the general category for anything to do with managing containers. Similarly `docker build` is now under `docker image build`. I don't see any plan to deprecate them yet but it's certainly the more up-to-date way to run containers.

xlr-8:
I'll check!

xlr-8:
> @xlr-8, the reason this is done is because the filename and s3_* parameters on the aws_lambda_function Terraform resource conflict with one another. The separate module takes care of detecting which parameters have been passed and provisions the Lambda function one way or the other and then passes back the ARN etc. of whichever one was used.

OK, I think I get it, however couldn't we have the outputs (`${element...}` to be used directly inside the first module), and same for the ternary on the `count`?

If I'm saying this, it's because it seems more complex to me to justify the usage of 2 modules, rather than sticking to 1 with eventually a comment if necessary. Let me know if you think this is feasible and/or if you agree/disagree.

Makefile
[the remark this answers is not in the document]
cristim: Fair enough, thanks for the clarification

grahamlyons:
Hey @xlr-8,

> couldn't we have the outputs (${element...} to be used directly inside the first module

The Terraform modules are analogous to functions so their outputs have to be within themselves. The reason the function has an output is because we need to get the name of the lambda resource back out but we don't know in advance which lambda resource will be used - it'll depend on whether a local file name or a location in S3 is passed in.

I think the neatest solution is to use another module here. As I say, we don't know in advance if we're going to get a local file or S3 object so we have to have some logic to switch between them (in this case the Terraform `count` trick).

We could possibly do it using multiple resources, with the same conditionals, at the same level as the existing lambda resource and use `locals` definitions instead of `output` variables from the module. To me that seems quite messy and I much prefer to encapsulate that logic into a module.

Let me know what you think.

xlr-8:
@grahamlyons 
> The Terraform modules are analogous to functions so their outputs have to be within themselves. The reason the function has an output is because we need to get the name of the lambda resource back out but we don't know in advance which lambda resource will be used - it'll depend on whether a local file name or a location in S3 is passed in.

Yes, I that's what I understood as well. I honestly think that it could be fine with locals, to me it's about moving the 2 lambda blocks, and outputs as locals; which is pretty much the same. I don't have strong opinions on that, what you did does the job just fine, and is clean; so I can't complain :)

Regarding the user, he already has to know already what he's using when calling the module with the proper argument (file / S3) otherwise he might end up in undefined variables I believe.

xlr-8:
I'll double check once the documentation has been update to re-validate that the documentation's sample are working fine.

Speaking of which I made a tiny gist to create a basic testing stack: https://gist.github.com/xlr-8/a33ab0eefad1adc6a4a8e44c9f04cc35 - not sure if I should document it or not.

Same for using S3 remote state file, instead of local one. That could be handy at some point.

grahamlyons:
@xlr-8, I've added documentation into the `START.md` file. I wasn't sure where it would fit into `TECHNICAL_DETAILS.md` so I've not added anything into there.

It looks like your `sample.tf` gist could be useful for acceptance testing a deployed Lambda. Do you have anything like that in place at the moment?

xlr-8:
Cool updates! Yeah, sorry maybe I should have clarified a bit better what I had in mind, I made a patch to ease things: https://gist.github.com/xlr-8/a899a3923fb6a0e841ab071492877bdf

Regarding the `sample.tf` it could indeed be used for such purpose, we currently do not have acceptance testing of a deployed Lambda, but that's how I'm checking simple code updates. I have thought of including to the repo for such purpose, but haven't taken the time quite yet.

grahamlyons:
@xlr-8, thanks for the patch. I've incorporated those changes so hopefully that's as you'd like it now.

cristim:
LGTM.

@xlr-8 feel free to merge this if you have no other comments.

xlr-8:
Thanks @grahamlyons for your work, time, and patience :)

PR #227, opened 2018-03-01
Use Security Group IDs in all cases

Shugyousha:
# Issue Type

- Bugfix Pull Request for https://github.com/cristim/autospotting/issues/48.

## Summary

This PR is one way to fix https://github.com/cristim/autospotting/issues/48.  (The first commit is just an unrelated cleanup commit however.)

We want to attach the correct SecurityGroups to newly created spot instances. For that we just reuse the SecurityGroup attached to the instances in the auto scaling group but it turns out that those Security Groups can either be mentioned by name (in the classic EC2 or the DefaultVPC case) or by id (in the Lunix/UNIX (VPC) case).

In order to make the handling of these different cases easier we try to use IDs in all cases. If a SecurityGroup string starts with "sg-" we assume it's an idea already and use it as is. Otherwise we search for the string as a group name using the AWS EC2 API and use its ID.

I think this should also cover the case where the security groups of the instance are mentioned in a mixed way, containing both names and IDs (as long as the names don't start with "sg-").

In our environment I can only test the DefaultVPC case so we have to make sure that this approach still works for all of the other cases.

### DIFF
--- a/core/launch_configuration.go
+++ b/core/launch_configuration.go
@@ -2,6 +2,7 @@ package autospotting
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -11,6 +12,8 @@ import (
 
 type launchConfiguration struct {
 	*autoscaling.LaunchConfiguration
+
+	secGroupRegex *regexp.Regexp
 }
 
 func (lc *launchConfiguration) countLaunchConfigEphemeralVolumes() int {
@@ -86,7 +89,7 @@ func (lc *launchConfiguration) convertLaunchConfigurationToSpotSpecification(
 		}
 	}
 
-	secGroupIDs, err := getSecurityGroupIDs(conn, lc.SecurityGroups)
+	secGroupIDs, err := lc.getSecurityGroupIDs(conn, lc.SecurityGroups)
 	if err != nil {
 		return nil, err
 	}
@@ -154,7 +157,7 @@ func copyBlockDeviceMappings(
 // We don't know whether we got security group names or ids. We assume
 // that the ones starting with "sg-" are ids and then search for the IDs
 // of the other ones.
-func getSecurityGroupIDs(conn *connections, secGroups []*string) ([]*string, error) {
+func (lc *launchConfiguration) getSecurityGroupIDs(conn *connections, secGroups []*string) ([]*string, error) {
 	var (
 		names    []*string
 		ids      []*string
@@ -163,8 +166,8 @@ func getSecurityGroupIDs(conn *connections, secGroups []*string) ([]*string, err
 	)
 
 	for _, secGroupStr := range secGroups {
-		// we assume these are IDs already
-		if strings.HasPrefix(*secGroupStr, "sg-") {
+		// we assume strings that match are IDs already
+		if lc.secGroupRegex.MatchString(*secGroupStr) {
 			ids = append(ids, aws.String(*secGroupStr))
 		} else {
 			names = append(names, aws.String(*secGroupStr))
--- a/core/launch_configuration.go
+++ b/core/launch_configuration.go
@@ -12,9 +12,9 @@ import (
 
 type launchConfiguration struct {
 	*autoscaling.LaunchConfiguration
-
-	secGroupRegex *regexp.Regexp
 }
+
+var secGroupRegex = regexp.MustCompile(`^sg-[a-f0-9]{8,17}$`)
 
 func (lc *launchConfiguration) countLaunchConfigEphemeralVolumes() int {
 	count := 0
@@ -163,7 +163,7 @@ func (lc *launchConfiguration) getSecurityGroupIDs(conn *connections, secGroups 
 
 	for _, secGroupStr := range secGroups {
 		// we assume strings that match are IDs already
-		if lc.secGroupRegex.MatchString(*secGroupStr) {
+		if secGroupRegex.MatchString(*secGroupStr) {
 			ids = append(ids, aws.String(*secGroupStr))
 		} else {
 			names = append(names, aws.String(*secGroupStr))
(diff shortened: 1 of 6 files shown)


cristim:
Thanks for your contribution, I'll look into it ASAP.

core/launch_configuration.go
```diff
@@ -85,33 +86,24 @@ func (lc *launchConfiguration) convertLaunchConfigurationToSpotSpecification(
 		}
 	}
 
+	secGroupIDs, err := getSecurityGroupIDs(conn, lc.SecurityGroups)
+	if err != nil {
+		return nil, err
```
cristim: I think this error condition could indicate we're running on EC2 Classic, where we can go ahead instead of returning here, and use the names as they are instead of using IDs.

I'd try to test this on one of our EC2-Classic environments to see what it actually happens.

core/launch_configuration.go
```diff
@@ -159,3 +150,43 @@ func copyBlockDeviceMappings(
...
+	var (
+		names    []*string
+		ids      []*string
+		outNames *ec2.DescribeSecurityGroupsOutput
+		err      error
+	)
+
+	for _, secGroupStr := range secGroups {
+		// we assume these are IDs already
+		if strings.HasPrefix(*secGroupStr, "sg-") {
```
cristim: Please make this check more strict, because the IDs have a certain format, fixed length and certain allowed characters so we could also use regex matching.

Shugyousha: Ok, I can use a regex to do that. Do you want me to add commits to this PR or just rebase my changes?

core/launch_configuration.go
Shugyousha: IIRC if we are on EC2-Classic, the securitygroups will be given as names. I would expect those to just be converted successfully to IDs here (if the API call is succesfull) and no error to be returned.

core/launch_configuration.go
cristim: Feel free to add more commits, I'll squash them all at merge time.

cristim:
Please also try to address the codeclimate issues: https://codeclimate.com/github/cristim/autospotting/pull/227

Shugyousha:
> Why do we need the regex here? We should just have test cases with actual group IDs set on the instance data structures.

Well, you wanted me to check it:
> Please make this check more strict, because the IDs have a certain format, fixed length and certain allowed characters so we could also use regex matching.

I will allow for the longer format in the regex and add some more test cases.

I am not convinced by those codeclimate issues. It will most likely be hard to fix because I don't know how they calculate their "Cognitive Complexity". I also think that having 5 return statements in a function is not an issue (keyword "early return").

cristim:
I wanted to define and check the regex in the code, the tests should just contain test cases with strings we expect the regex to match or not and not re-define the regex once again.

The codeclimate issues are sometimes questionable. We can accept them if there's no other way but I'd like to at least give it a try to simplify the code, if possible with reasonable effort.

Shugyousha:
I fixed one codeclimate issue and improved the other two.

core/launch_configuration_test.go
```diff
@@ -106,7 +106,7 @@ func Test_copyBlockDeviceMappings(t *testing.T) {
 	}
 }
 
-var testSecGroupRegex = regexp.MustCompile(`^sg-[a-f0-9]{8}$`)
+var testSecGroupRegex = regexp.MustCompile(`^sg-[a-f0-9]{8,17}$`)
```
cristim: Can't we delete this from here? It should be enough if we just use what's defined in core/autoscaling.go:850

Shugyousha: We need access to the same regex in the tests as well. What we could do is to define the regex as a global in autoscaling.go and then use that one in both getLaunchConfiguration() and the tests in launch_configuration_test.go. Would you prefer that?

cristim: But why do we need it?

The tests should just hit the actual code that has the regex, unless I'm missing something.

Shugyousha:
On Fri, Mar 2, 2018 at 3:27 PM, Cristian Măgherușan-Stanciu
@magheru_san <notifications@github.com> wrote:
> @cristim commented on this pull request.
>
> ________________________________
>
> In core/launch_configuration_test.go:
>
>> @@ -106,7 +106,7 @@ func Test_copyBlockDeviceMappings(t *testing.T) {
>   }
>  }
>
> -var testSecGroupRegex = regexp.MustCompile(`^sg-[a-f0-9]{8}$`)
> +var testSecGroupRegex = regexp.MustCompile(`^sg-[a-f0-9]{8,17}$`)
>
> But why do we need it?
>
> The tests should just hit the actual code that has the regex, unless I'm
> missing something.

No, the regex is only initialised in getLaunchConfiguration() and we
don't call that to create the expectedLC in the tests. So in the
expectedLC the regex is actually a nil pointer if we don't initialise
it and DeepEqual seems to complain about that (since 'lc' with which
we compare it contains an initialised regexp and is thus not equal).

cristim:
I now understand what you mean.

But why is the regex a member of the launchConfiguration struct at core/launch_configuration.go:16 in the first place?

I would just define it as a stand-alone constant at the top of the file.

Shugyousha:
On Fri, Mar 2, 2018 at 4:22 PM, Cristian Măgherușan-Stanciu
@magheru_san <notifications@github.com> wrote:
> I now understand what you mean.
>
> But why is the regex a member of the launchConfiguration struct at
> core/launch_configuration.go:16 in the first place?
>
> I would just define it as a stand-alone constant at the top of the file.

I did not want to pollute the namespace with a global var.

I will put it outside and use it in getSecurityGroupIDs function. Then
I can remove it from the tests as well. Does that sound ok?

cristim:
That's exactly what I meant, thanks!

Shugyousha:
On Fri, Mar 2, 2018 at 4:48 PM, Cristian Măgherușan-Stanciu
@magheru_san <notifications@github.com> wrote:
> That's exactly what I meant, thanks!

Done! Saves us quite a few lines which is nice.

xlr-8:
Thanks for the code contribution, as this have been merged, I believe we could update the documentation to remove the bug link: https://github.com/cristim/autospotting/blob/master/TECHNICAL_DETAILS.md#summary-of-supported-features--options (at 'Support AWS EC2Classic') WDYT?

cristim:
I haven't been able to test it on EC2 Classic yet, let's only delete it once it is successfully tested

Issue #94, opened 2017-03-09
Remove requirement to run in us-east-1


phils:
## Issue type ##

- Feature Idea

## Summary ##

The docs say:

> For technical reasons the stack needs to be launched in the US-East-1(Virginia) region, so make sure it's not created in another region.

Would be good to be able to run this in other regions where requirements dictate this.  Can you elaborate on the technical reason it must run in us-east-1 at the moment?  Thanks.

cristim:
It's a CloudFormation limitation: when creating Lambda functions with CloudFormation it requires that the S3 bucket used for the function source code is located in the same region as the stack:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html

The bucket where TravisCI deploys the code after each build is located in US-East-1, so that's why the stack is must be created there as well.

If you want to use CloudFormation in a different region you will need to self-host your binaries, which means you need to take care of building the code and pushing it to your own S3 bucket, which has to be in that region.

As far as I know, the Terraform deployment code doesn't have this requirement, but I didn't try that yet.

xlr-8:
I confirm the limitation isn't present in terraform, because no S3 bucket is being used.
The zip archive is being sent directly rather than stored on S3 like it is the case for Cloudformation.

It uses the `filename` parameter [1]: https://www.terraform.io/docs/providers/aws/r/lambda_function.html#filename. So as documented, an S3 bucket could be used, but as it could have been avoided I wanted to keep things simple & stupid.
Therefor the region being used would be the one specified in the CLI: https://github.com/cristim/autospotting/blob/master/START.md#install-via-terraform

[1]. https://github.com/cristim/autospotting/blob/master/terraform/autospotting/lambda.tf#L3

cristim:
I think we can close this, since this is not a limitation of AutoSpotting but of the CloudFormation deployment mechanism when used against the provided TravisCI builds. When running a self-built or self-hosted S3 bucket (you should be able to `s3 sync` my bucket into yours), as well as when deploying with Terraform, the users don't have this limitation.

This is a stupid limitation of CloudFormation which I have no plans to mitigate, I won't provision the official builds S3 bucket in every single AWS region.

@phils If this is important to you please raise this to AWS, once they fix it in CloudFormation it should work out of the box.

phils:
[...]

PR #232, opened 2018-03-09
Add CloudFormation region note

brennv:
## Issue Type
- Documentation Pull Request

## Summary

Adds note as suggested in this [issue](https://github.com/cristim/autospotting/issues/94#issuecomment-286696488) for CloudFormation users attempting to launch outside us-east-1.

   to it.
   and
   configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,9 @@ For more detailed information you can read this [document](START.md)
 
 [![Launch](https://s3.amazonaws.com/cloudformation-examples/cloudformation-launch-stack.png)](https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/new?stackName=AutoSpotting&templateURL=https://s3.amazonaws.com/cloudprowess/nightly/template.json)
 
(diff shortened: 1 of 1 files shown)


xlr-8:
Thanks, I would have added something like: `:warning:` (:warning:) to make it more 'eye-catching' as it's some limitation

xlr-8:
@brennv do you have a minute, otherwise I can take care of this?

coretemp:
@cristim You can deploy it to other regions by using https://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html.

cristim:
I would still have to keep some 15 buckets în sync, and pay for the storage costs, for very little benefit. Speaking of, din anyone measure the impact of the current situation to the monthly runtime costs? 

For the stable build users I'm going to do it upon demand, as part of the setup support.

cristim:
Now that we use Docker images for the Lambda, the code should now run in any region, so this is no longer relevant.

Issue #138, opened 2017-08-29
 detachAndTerminateOnDemandInstance should use TerminateInstanceInAutoScalingGroup


cristim:
# Github issue #

## Issue type ##

- Feature Idea

## Summary ##

The current logic issues a detach followed by a termination.
`TerminateInstanceInAutoScalingGroup` would do these for us, slightly simplifying the replacement logic.

francoran:
@cristim I think that if we'll replace the detach logic then connection draining will not take place as it sends termination request to the instance. I didn't test it

cristim:
@francoran can you also test the patch in #141? I'd like to have that merged but I can't do it before it's confirmed to work as expected and for now I have no bandwidth to thoroughly test this.

phils:
@cristim  @francoran I use TerminateInstanceInAutoScalingGroup (via awscli) all the time in production and it definitely does ELB connection draining.

xlr-8:
I'm reopening this issue based on https://github.com/cristim/autospotting/pull/241/files

cristim:
Fixed in #287, closing issue

PR #241, opened 2018-04-08
Add a sleep to let detachment initialize properly before terminating

atillamas:
# Issue Type

- Bugfix Pull Request

## Summary

When detaching an instance from a Autoscaling Group autospotting just starts the detach and then immediately terminate the instance, this will make some requests during this time fail:
To reproduce: create an ALB, put 2 instances with a simple webserver for example  [http-echo](https://github.com/hashicorp/http-echo) and launch 2 new instances, result will be some missing requests when autospotting swap the on-demand instance to  spot-instances

```

Sun Apr 8 14:12:23 CEST 2018 version4: ip-10-0-62-37
Sun Apr 8 14:12:23 CEST 2018<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
</body>
</html>
Sun Apr 8 14:12:23 CEST 2018<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
</body>
</html>
Sun Apr 8 14:12:23 CEST 2018 version4: ip-10-0-62-37

```
It probably would be better to evaluate the detachment process in some way, but as a quick fix this should be sufficient. I don't know what you prefer?

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -917,6 +917,9 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 		return err
 	}
 
+	// Wait till detachment is complete before terminate instance
+	time.Sleep(20 * time.Second)
+
 	return a.instances.get(*instanceID).terminate()
 }
 
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -918,7 +918,7 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 	}
 
 	// Wait till detachment initialize is complete before terminate instance
-    time.Sleep(20 * time.Second * s.region.conf.SleepMultiplier)
+    time.Sleep(20 * time.Second * a.region.conf.SleepMultiplier)
 
 	return a.instances.get(*instanceID).terminate()
 }
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -917,7 +917,7 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 		return err
 	}
 
-	// Wait till detachment is complete before terminate instance
+	// Wait till detachment initialize is complete before terminate instance
 	time.Sleep(20 * time.Second)
 
 	return a.instances.get(*instanceID).terminate()
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -918,7 +918,7 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 	}
 
 	// Wait till detachment initialize is complete before terminate instance
-    time.Sleep(20 * time.Second * a.region.conf.SleepMultiplier)
+	time.Sleep(20 * time.Second * a.region.conf.SleepMultiplier)
 
 	return a.instances.get(*instanceID).terminate()
 }
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -918,7 +918,7 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 	}
 
 	// Wait till detachment initialize is complete before terminate instance
-	time.Sleep(20 * time.Second)
+    time.Sleep(20 * time.Second * s.region.conf.SleepMultiplier)
 
 	return a.instances.get(*instanceID).terminate()
 }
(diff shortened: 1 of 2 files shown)


cristim:
Is your ALB configured with the connection draining enabled? That should wait for your terminating instances.

The sleep may be an ugly hack, but we could tag and let the instance running until the next autospotting run.

atillamas:
@cristim Yes, I've tried with everything from 300 seconds draining to 0, makes no difference, I cannot reproduce the behavior manually, it's only when autospotting calls detach and then immediately after terminate. When adding the delay between the detach and terminate i get no lost requests at all.

Maybe it would be better to check if the detach is registered via this (this is the aws-cli reference, don't know how it would be done with go) 
`autoscaling describe-scaling-activities --auto-scaling-group-name testing-nomad-consul-client-t2.large-spot --query 'Activities[]`

cristim:
@atillamas that sounds strange. Try patching the code with the sleep you propose and see what happens, although that's hackish and I wouldn't do it in the mainline build.

The best long term solution would be to detach in the current run, mark for termination and only actuallyt terminate in the next. Would you be willing to make a PR for this kind of change?

atillamas:
@cristim I have patched and tested it, and it does work, the problem dissapears, otherwise i wouldn't have submitted it as a PR :)

I could give it a try to do as you suggested with the detach and then terminate in the next run.

cristim:
If the sleep improves the current situation I can approve it for now, although I don't know how this can help for longer running connections that would be open for longer than the sleep interval. Did you test this condition?

But for long term I think I prefer the next-run termination. It would be great if you could contribute that.

core/autoscaling.go
```diff
@@ -917,6 +917,9 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
 		return err
 	}
 
+	// Wait till detachment initialize is complete before terminate instance
+	time.Sleep(20 * time.Second)
```
cristim: Please also use the  SleepMultiplier so that we don't actually sleep when running the tests.
This would become`time.Sleep(20 * time.Second * s.region.conf.SleepMultiplier)`

atillamas:
@cristim Sorry, don't understand what you mean with the longer running connection? The sleep doesn't actually wait for the detach to finish, it only waits for it to initialize, Since the node is set to draining state it doesn't matter if it's terminated directly after, It's just that without it, the node never gets the draining state correctly  in the target-group before it's terminated. Seems like aws is doing some decoupling when terminating? I don't actually know what.

cristim:
I see what you mean, it's the eventual consistency that we also noticed when trying to tag newly created spot instance requests. We used to see errors because the resources were present but not yet ready to be tagged.

atillamas:
@cristim Sorry, no i haven't tested for long running connections, It doesn't seem like it's taken into account now anyway.
They would have to reconnect no matter what. Otherwise we would have to wait forever?
I guess it will be better if we check if the node is completely drained from the target group and then swap it out.

core/autoscaling.go
cristim: I did a typo: should likely be `a.region.conf.SleepMultiplier`

atillamas: Fixed

atillamas:
@cristim Yes, it's the `* a.region.conf.SleepMultiplier` part that you requested that breaks  it, don't know how to fix ut unfortunately.

cristim:
Sorry, I'll fix this ASAP.

cristim:
I fixed it in https://github.com/cristim/autospotting/commit/5392a04782df70df734948f81855a0d8c0a6716c, please cherry-pick that and update the PR.

atillamas:
@cristim Thank you! Updated the PR

xlr-8:
> @xlr-8 what's your take on this?

It is that we had a [PR](https://github.com/cristim/autospotting/pull/141) for actually draining connections properly, but that was reverted due to an issue you mentioned [here](https://github.com/cristim/autospotting/commit/28ede473ffb42147823f4085703d7961af23201f).

So I'm not surprised that we have an issue with the draining of connection, but as you commented the sleep is hacky, though 20sec can be quite a lot. To try making it cleaner, I would have:
* Tried to use/create a method for asg/instances to know if the one we are speaking about is still in the ASG/ELB (still with the sleep/timeout logic)
* Have a proper fix with actually calling `TerminateInstanceInAutoScalingGroup` as it was done before but without issues

Due to that I reopened https://github.com/cristim/autospotting/issues/138

xlr-8:
@atillamas If you were willing to work on https://github.com/cristim/autospotting/issues/138 to have a long-term fix that be kind of you!

atillamas:
@xlr-8 I could give it a shot

Issue #242, opened 2018-04-12
Implement a opt-out mode for the AutoScaling group tag filters


cristim:
# Github issue #

## Issue type ##

- Feature Idea

## Summary ##

Currently AutoSpotting works in opt-in mode, only handling the groups tagged with spot-enabled=true or another tag configured by the user. In some cases, it may be desirable to run in opt-out mode, running on all groups unless they are tagged with an opt-out tag value.

## Proposal ##
 
- Implement an additional, opt-out mode, that runs against all groups by default.
- Allow opt-out on certain groups based on the same spot-enabled tag but when set to false
- Make the tag configurable in a similar way to how we did it for the opt-in mode so users can pass any other values.
- Implement a selector for the opt-in/opt-out modes, defaulting to the current opt-in behavior.

PR #244, opened 2018-04-20
Add support for opt-out mode

cristim:
Implements #242.

# Issue Type

- Feature Pull Request

## Summary

Implements #242, see that for more details.

   to it.
   and
   configurations.

### DIFF
--- a/core/main.go
+++ b/core/main.go
@@ -27,6 +27,7 @@ func Run(cfg *Config) {
 	// use this only to list all the other regions
 	ec2Conn := connectEC2(cfg.MainRegion)
 
+	addDefaultFilteringMode(cfg)
 	addDefaultFilter(cfg)
 
 	allRegions, err := getRegions(ec2Conn)
@@ -40,9 +41,20 @@ func Run(cfg *Config) {
 
 }
 
+func addDefaultFilteringMode(cfg *Config) {
+	if cfg.TagFilteringMode == "" {
+		cfg.TagFilteringMode = "opt-in"
+	}
+}
+
 func addDefaultFilter(cfg *Config) {
 	if len(strings.TrimSpace(cfg.FilterByTags)) == 0 {
-		cfg.FilterByTags = "spot-enabled=true"
+		switch cfg.TagFilteringMode {
+		case "opt-out":
+			cfg.FilterByTags = "spot-enabled=false"
+		default:
+			cfg.FilterByTags = "spot-enabled=true"
+		}
 	}
 }
 
--- a/core/main.go
+++ b/core/main.go
@@ -42,8 +42,12 @@ func Run(cfg *Config) {
 }
 
 func addDefaultFilteringMode(cfg *Config) {
-	if cfg.TagFilteringMode == "" {
+	if cfg.TagFilteringMode != "opt-out" {
+		debug.Printf("Configured filtering mode: '%s', considering it as 'opt-in'(default)\n",
+			cfg.TagFilteringMode)
 		cfg.TagFilteringMode = "opt-in"
+	} else {
+		debug.Println("Configured filtering mode: 'opt-out'")
 	}
 }
 
--- a/core/main_test.go
+++ b/core/main_test.go
@@ -128,6 +128,13 @@ func Test_addDefaultFilterMode(t *testing.T) {
 			},
 			want: "opt-out",
 		},
+		{
+			name: "Anything else gives the opt-in FilterMode",
+			cfg: Config{
+				TagFilteringMode: "whatever",
+			},
+			want: "opt-in",
+		},
 	}
 
 	for _, tt := range tests {
--- a/core/main_test.go
+++ b/core/main_test.go
@@ -84,6 +84,11 @@ func Test_spotEnabledIsAddedByDefault(t *testing.T) {
 			},
 			want: "spot-enabled=true",
 		},
+		{
+			name:   "Default No ASG Tags",
+			config: Config{TagFilteringMode: "opt-out"},
+			want:   "spot-enabled=false",
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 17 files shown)


core/main.go
```diff
@@ -40,9 +41,20 @@ func Run(cfg *Config) {
 
 }
 
+func addDefaultFilteringMode(cfg *Config) {
+	if cfg.TagFilteringMode == "" {
```
xlr-8: I would change that method to make it as such:

core/main.go
```diff
@@ -40,9 +41,20 @@ func Run(cfg *Config) {
...
+		cfg.TagFilteringMode = "opt-in"
+	}
```
xlr-8: I would change that method to make it as such:
```go
if cfg.TagFilteringMode != "opt-out" {
```

This is because there are two valid options: `opt-in` (default) and `opt-out`. But with the current method, the validation of `cfg.TagFilteringMode = "whatever"` would result in weird behaviour; while with this change anything not being valid would result in `opt-in`. You could also add a debug/warning message for the users; or even stop the execution, but that's a bit harsh.

core/main_test.go
```diff
@@ -97,3 +102,41 @@ func Test_spotEnabledIsAddedByDefault(t *testing.T) {
...
+			},
+			want: "opt-in",
+		},
+		{
+			name: "Opt-out FilterMode",
+			cfg: Config{
+				TagFilteringMode: "opt-out",
+			},
+			want: "opt-out",
+		},
```
xlr-8: With the previous comment, you could now add the test of an invalid input as well such as `whatever`, and you'd still want `opt-in`.

xlr-8: With the previous comment, you could now add the test of an invalid input as well such as `whatever`, and you'd still want `opt-in`.

core/region.go
```diff
@@ -330,22 +330,30 @@ func isASGWithMatchingTags(asg *autoscaling.Group, tagsToMatch []Tag) bool {
...
+		// Go lacks a logical XOR operator, this is the equivalent to that logical
+		// expression. The goal is to add the matching ASGs when running in opt-in
+		// mode and the other way round.
+		if optInFilterMode != groupMatchesExpectedTags {
+			continue
 		}
-	}
 
+		asgName := *group.AutoScalingGroupName
+		logger.Println("Matching tags found for ASG, enabling ASG for processing:", asgName)
```
xlr-8: This message could now be a bit miss-leading, considering that in the `opt-out`, the lack of tag would be the reason for which they got selected.

Maybe using: `Tag (%s) and mode (%s) matching of this ASG` could do? Thus making explicit the tag/mode.

core/region_test.go
```diff
@@ -435,6 +435,50 @@ func TestFilterAsgs(t *testing.T) {
...
+										{Key: aws.String("environment"), Value: aws.String("qa"), ResourceId: aws.String("asg4")},
+									},
+									AutoScalingGroupName: aws.String("asg4"),
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
xlr-8: May you add some tests with "custom tags" as well? Now that we have implemented the `-tag_filters` option, it would be great to ensure that it works fine with the opt-out mode, whether this is one or multiple tags actually as stated in the help `'spot-enabled=true,Environment=dev,Team=vision'`.

terraform/autospotting/variables.tf
```diff
@@ -47,7 +49,24 @@ variable "autospotting_regions_enabled" {
 }
 
 variable "autospotting_tag_filters" {
-  description = "tags to filter which ASGs autospotting considers.  by default if left blank, asgs tagged with spot-enbled=true will be operated on.  You can set this to many tags, for example: spot-enabled=true,Environment=dev,Team=vision"
+  description = <<EOF
+  Tags to filter which ASGs autospotting considers. By default, ASGs tagged
+  with spot-enbled=true will be operated on. In opt-out mode it operates on
+  all groups except those tagged with spot-enbled=false.
```
xlr-8: I think that the tags for opt-out should be also an input; so that people could simply opt-out all `dev` ASGs for example. Otherwise, we might run in the same case previously established for `spot-enabled=true`

terraform/variables.tf
```diff
@@ -55,8 +66,26 @@ variable "asg_regions_enabled" {
...
-  description = "Tags to filter which ASGs autospotting considers.  If blank by default this will search for asgs with spot-enabled=true.  You can set this to many tags, for example: spot-enabled=true,Environment=dev,Team=vision"
-  default     = ""
+  description = <<EOF
+  Tags to filter which ASGs autospotting considers. If blank
+  by default this will search for asgs with spot-enabled=true (when in opt-in
+  mode) and will skip those tagged with spot-enabled=false when in opt-out mode.
+
+  You can set this to many tags, for example:
+  spot-enabled=true,Environment=dev,Team=vision
+  EOF
```
xlr-8: Same comment as previously

terraform/autospotting/variables.tf
cristim: Not sure I understand your comment.

By design this field should be used in both opt-in and opt-out mode and its behavior is only controlled by the autospotting_tag_filtering_mode variable.

core/main_test.go
cristim: Done

cristim: Done

core/region_test.go
cristim: Done

cristim:
I addressed most comments, except for those which I'm not sure I understand.
@xlr-8 can you explain those on gitter?

cristim:
I decided to merge this, it was sitting for too long in review.

terraform/autospotting/variables.tf
xlr-8: My understanding when I first read the PR was that the opt-out, would only consider `spot-enabled=false` and no other tags. For example, we might have wanted to opt-out all production environments ASGs.

Maybe I got it wrong, but I feel that it would have been useful, if that wasn't the case.

xlr-8:
Yep, sorry I had issues to get some time out of my schedule this week.

terraform/autospotting/variables.tf
cristim: It will consider all the tags configured in the tag filter list. But when none is configured it takes the `spot-enabled=false`

PR #245, opened 2018-04-20
Implement tagging for the Lambda function

cristim:
# Issue Type

- Feature Pull Request

## Summary
This makes the CloudFormation stack more human-friendly by converting it to YAML, and implements tagging for the Lambda function in both CloudFormation and Terraform.

You can see #247 for more context.

   to it.
   and
   configurations.


cristim:
Still todo: make sure the docs are pointing to the YAML stack code.

cloudformation/stacks/AutoSpotting/parameters.yaml
```diff
@@ -1,17 +1,9 @@
...
+LambdaZipPath: nightly/lambda.zip
 MinOnDemandNumber: '1'
-
-# Valid choices:
-# - for EC2 Classic environments: Linux/UNIX | SUSE Linux | Windows
-# - for VPC: Linux/UNIX (Amazon VPC) | SUSE Linux (Amazon VPC) | Windows (Amazon VPC)
-SpotProductDescription: "Linux/UNIX (Amazon VPC)"
\ No newline at end of file
+SpotProductDescription: Linux/UNIX (Amazon VPC)
+TagFilteringMode: opt-out
```
xlr-8: Why removing the comments?

xlr-8: Why removing the comments though?

cloudformation/stacks/AutoSpotting/template.yaml
```diff
@@ -0,0 +1,262 @@
...
+        State: "ENABLED"
+        Targets:
+          -
+            Arn:
+              Fn::GetAtt:
+                - "LambdaFunction"
+                - "Arn"
+            Id: "AutoSpottingEventGenerator"
+      Type: "AWS::Events::Rule"
+
```
xlr-8: As JSON is a subset of YAML, I assume you made an automatic conversion?

cloudformation/stacks/AutoSpotting/parameters.yaml
cristim: It was done automatically by the tool that I used to convert the stack code.

I will bring them back.

cloudformation/stacks/AutoSpotting/template.yaml
cristim: Yes, I used `json2yaml template.json > template.yaml`

PR #248, opened 2018-04-27
use TerminateInstanceInAutoScalingGroup instead of Detach

### LINKED ISSUE #138: detachAndTerminateOnDemandInstance should use TerminateInstanceInAutoScalingGroup

jburnham:
# Issue Type
- Feature Pull Request
## Summary

Re-introducing and re-basing the change from #141  that replaces the detach then terminate workflow with the single api call [TerminateInstanceInAutoScalingGroup](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_TerminateInstanceInAutoScalingGroup.html). This patch is running in our custom build of autospotting and has been successfully replacing our prod machines with spot instances consistently and without reproducing the issues that were [previously mentioned](https://github.com/cristim/autospotting/pull/141#issuecomment-339091432). The machines in our ASG scale up and down throughout the day so autospotting has to work as well and it maintains full spot coverage shortly after scale-out actions.

The TerminateInstanceInAutoScalingGroup call is allowing our machines in the ASG to execute their Terminate lifecycle hook before actual termination. This hook tells the machine to start draining ECS tasks which causes them to move the tasks to the spun up spot instance so there is minimal disturbance to our applications. If users had these machines in an [ELB](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-conn-drain.html)/[ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#deregistration-delay), they should do connection draining as well. I confirmed and tested with an instance in an ALB and it stayed in the Termination step until the ALB finished the de-registration delay before the machine was actually shutdown.

Fixes #138 

As this change was previously reverted I would be very happy to help troubleshooting any issues with this patch. Looking at the code I can't see an obvious reason why things would go into a loop of spinning up and then terminating spot instances. More information on maybe how the ASG is setup and the state of it when the loops happens might help.

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -6,7 +6,6 @@ import (
 	"math"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -598,15 +598,15 @@ func (a *autoScalingGroup) attachSpotInstance(spotInstanceID string) error {
 	return nil
 }
 
-// Terminates an on-demand instance from the group,
-// but only after it was detached from the autoscaling group
+// Terminates an on-demand instance from the group using the
+// TerminateInstanceInAutoScalingGroup api call.
 func (a *autoScalingGroup) terminateInstanceInAutoScalingGroup(
 	instanceID *string) error {
 	logger.Println(a.region.name,
 		a.name,
-		"Detaching and terminating instance:",
+		"Terminating instance:",
 		*instanceID)
-	// detach the on-demand instance
+	// terminate the on-demand instance
 	terminateParams := autoscaling.TerminateInstanceInAutoScalingGroupInput{
 		InstanceId:                     instanceID,
 		ShouldDecrementDesiredCapacity: aws.Bool(true),
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -618,8 +618,6 @@ func (a *autoScalingGroup) terminateInstanceInAutoScalingGroup(
 		return err
 	}
 
-	// Wait till detachment initialize is complete before terminate instance
-	time.Sleep(20 * time.Second * a.region.conf.SleepMultiplier)
 	return nil
 }
 
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -289,7 +289,7 @@ func (a *autoScalingGroup) needReplaceOnDemandInstances() bool {
 		logger.Println("Currently OnDemand running equals to the required number, skipping run")
 		return false
 	}
-	logger.Println("Currently less OnDemand instances than required !")
+	logger.Println("Currently fewer OnDemand instances than required !")
 	if a.allInstanceRunning() && a.instances.count64() >= *a.DesiredCapacity {
 		logger.Println("All instances are running and desired capacity is satisfied")
 		if randomSpot := a.getAnySpotInstance(); randomSpot != nil {
@@ -298,7 +298,7 @@ func (a *autoScalingGroup) needReplaceOnDemandInstances() bool {
 			} else {
 				logger.Println("Terminating a random spot instance",
 					*randomSpot.Instance.InstanceId)
-				randomSpot.terminate()
+				a.terminateInstanceInAutoScalingGroup(randomSpot.Instance.InstanceId)
 			}
 		}
 	}
@@ -433,7 +433,7 @@ func (a *autoScalingGroup) replaceOnDemandInstanceWithSpot(
 		defer a.attachSpotInstance(spotInstanceID)
 	}
 
-	return a.detachAndTerminateOnDemandInstance(odInst.InstanceId)
+	return a.terminateInstanceInAutoScalingGroup(odInst.InstanceId)
 }
 
 // Returns the information about the first running instance found in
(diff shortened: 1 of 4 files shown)


cristim:
@jburnham Thanks for reviving this work, I agree this is a really valuable contribution and I'd love to have it finally merged once we can test it and confirm that we have no regressions.

I looked at the code and it seems not much changed since last time, and it's a bit strange, I have no idea what caused the spinning last time we tried it.

I'll try to test it over the next few days and get back to you with my findings.

tootedom:
Hi there,
Is it possible to have a configuration option for this, and keep the default as is? 
We use the termination hooks of the asg, currently, to only notify us of an instance termination.
If we get many of them for the same app time, it is an indication that the app potentially has an issue (i.e. memory leak and crashes).

If we change the default to not be detach, then we are going to getting a lot of notifications.

Would it be possible to have a tag on the asg that controls whether autospotting:

- detaches and terminates
Or
- terminates in the asg

Default being the current behaviour (what people using autospotting are currently familiar with)

Would this be possible?
Thanks
/Dom

core/autoscaling.go
```diff
@@ -893,34 +893,27 @@ func (a *autoScalingGroup) attachSpotInstance(spotInstanceID *string) error {
...
-	// detach the on-demand instance
-	detachParams := autoscaling.DetachInstancesInput{
-		AutoScalingGroupName: aws.String(a.name),
-		InstanceIds: []*string{
-			instanceID,
-		},
+	// terminate the on-demand instance
+	terminateParams := autoscaling.TerminateInstanceInAutoScalingGroupInput{
+		InstanceId:                     instanceID,
 		ShouldDecrementDesiredCapacity: aws.Bool(true),
```
cristim: Did you test this against groups currently running at minimum capacity?

Will the termination still happen if the capacity can't be decreased because it's already at minimum?

jburnham: I just tested it and it worked. Right before the terminate action, it's attaching the spot instance so it should be 1 above the minimum count.
```
2018/04/30 11:42:18 autoscaling.go:313: us-west-2 Attaching spot instance i-0b42ef9ec5d5f6c0d to ecs
2018/04/30 11:42:18 autoscaling.go:424: ecs Retrieving instance details for  i-0b42ef9ec5d5f6c0d
2018/04/30 11:42:18 autoscaling.go:431: ecs i-0b42ef9ec5d5f6c0d is in the availability zone us-west-2b looking for an on-demand instance there
2018/04/30 11:42:18 autoscaling.go:444: ecs found on-demand instance i-010f042d7a7968a0c replacing with new spot instance i-0b42ef9ec5d5f6c0d
2018/04/30 11:42:18 autoscaling.go:900: us-west-2 ecs Terminating instance: i-010f042d7a7968a0c
```

Minimum and desired are at 5. After running the lambda, the number of instances goes to 6 from 5, while minimum and desired are still at 5. The terminating instance now executes its lifecycle hook as normal.

tootedom:
Hi there,

Please can we have a configuration option for this, and keep the default behaviour as is.  Otherwise, we will effect people's currently running expectations of how instances are removed from the asg and replaced with spot instances.  (Per my previous comment)

Detaching and termination are two separate things.  As we mention the pull request is about making it terminate in the asg so lifecycle hooks are executed.  Detaching ensures these are not run.  The current behaviour is the later.  Changing the default behaviour could have a nasty surprise for users (If they are unaware of the change).  A config option would be welcome for this behaviour change.  Is this possible?

Thanks
/Dom

otterley:
I disagree with you @tootedom.  

First, termination notifications are orthogonal to Lifecycle Hooks.  You can have a termination notification without a Lifecycle Hook, and you can have a Lifecycle Hook without a termination notification (although the latter is practically useless).

Second, sending alerts on Auto Scaling termination notifications is not a best practice.  There are lots of reasons why an Auto Scaling instance could be terminated, to the point where it's a best practice to consider it completely normal and expected.  There are plenty of other ways to monitor the health of an application; to base your detection solely on the termination of Auto Scaling instances would lead to too many false negatives.

I'd prefer that this software use best practices out of the box; people should be encouraged by it to design their systems well.

otterley:
@cristim are there any remaining concerns about this PR?  If not, can you please merge it?

cristim:
@otterley My main concern about this was caused by the regression that seemingly caused instances to get launched in a loop. This may have been a coincidence and actually caused by another bug, but I would like to have more people test it to make sure we don't reintroduce the same regression twice.

Regarding the behavior change, I would also prefer to do the right thing by default, but I understand that there are users out there who may rely on the current behavior, so we should do our best not to break their setup or at least to offer them a way to get the current behavior.

My proposal is to merge this change as it is (once rebased on the current code), but with some additional documentation about the behavior change, and then I would gladly accept a patch that introduces an additional configuration flag that allows users to get the current behavior back if they really need it.

Users who rely on the current behavior can be advised to keep running a previous version until that configuration option is added, or even invited to contribute this piece of functionality.

jburnham:
I have rebased onto the latest master. This PR is also flagged to accept maintainer edits.

tootedom:
Apologies to others, I know i am replying late and that the prior conversations have addressed.the comments.  However, as there were addressed at myself, I thought I would reply.

@otterley i agree with the change in principle. However, what I disagree with is the change to the default behaviour without an option for existing users to continue to support the current behaviour.  Regardless about about how any of us believe how 'hooks' should be used.  The software used to work 'x' way, now it will work 'y' way.   It is not up to me how to support the change.  All I was suggesting is a way to support either the old operation, or a clear documentation alert on how things will change  I.e the new release introduces a change that does X.  

When I get the time to merge master, etc.  I may or may not open or pull request to support the previous behaviour. It is how it is; and it was a valid suggestion on how to support changes in behaviour to software products.

Cheers
/dom

otterley:
100% agreed @tootedom.  I'll even code it up.

cristim:
Closing.

This was superseded by #287, which also implements configurable support for the previous instance termination behavior.

PR #256, opened 2018-05-22
Rebase 2

ktruckenmiller:
# Issue Type

- Feature Pull Request
- Bugfix Pull Request
- Documentation Pull Request

   to it.
   and
   configurations.


ktruckenmiller:
Accident, please close and delete

Issue #183, opened 2017-12-07
Investigate and document memory leaks and resource requirements for large installations


cristim:
# Github issue #

## Issue type ##

- Bug
- Documentation fix

## Build number #

nightly-468

## Configuration ##

any

## Environment ##

Large AWS accounts, with > 500 instances per region

## Summary ##

On large AWS accounts, where API throttling is quite common the Lambda function has been found to hang and timeout runs.

## Steps to reproduce ##

- Enable it on such a large account
- Trigger API throttling by performing many API calls

## Expected results ##

AutoSpotting should gracefully handle this situation

## Actual results ##

- The Lambda run timed out, seemed to hang after `Processing page 1 of DescribeInstancesPages`
- Lambda was running out of the allocated memory
- Spot Instances were launched but not tagged properly so they failed to be added to the group
- Increasing the memory allocation seemed to help, 3GB Lambda seemed to work fine.

symonds:
Account which is causing issues has:

400-500 Instances
258 ASG's of which 54 are tagged `spot-enabled = true`, the remaining ASG's are not tagged `spot-enabled` at all
Each ASG in the test env which has autospotting is enabled is min/max/desired = 1 :. 54 Instances should be converted from on-demand to spot
`allowed-instance-types` tag is blank
Instance sizes are a mixture of t2.small, m4.large, r4.large

xlr-8:
As discussed on gitter @symonds reported the following output - which confirmed my thought:
```bash
PrivateIpAddress: "10.6.0.180" PrivateIpAddress: "10.6.0.180" }], SourceDestCheck: false, Status: "in-use",
SubnetId: "subnet-73a1fb2a",
VpcId: "vpc-d67c4fb3"
}],
Placement: {
AvailabilityZone: "eu-west-1a",
GroupName: "",
Tenancy: "default"
},
PrivateDnsName: "ip-10-6-0-180.eu-west-1.compute.internal",
PrivateIpAddress: "10.6.0.180",
ProductCodes: [{
ProductCodeId: "aw0evgkw8e5c1q413zgy5pjce",
ProductCodeType: "marketplace"
}],
PublicDnsName: "<snip>.
PublicIpAddress: "<snip>",
RootDeviceName: "/dev/sda1",
RootDeviceType: "ebs",
SecurityGroups: [{
GroupId: "sg-0ce7bf68",
GroupName: "cog-test-vpn"
},{
GroupId: "sg-7ee7bf1a",
[...]

PR #257, opened 2018-05-25
Switch to the RunInstances API call

cristim:
# Issue Type

- Feature Pull Request

## Summary

This makes use of the relatively new capabilities of the RunInstances API call, which can now launch spot instances by setting a few extra fields in addition to what is needed to launch an on-demand instance.

With this we no longer need to perform and keep track of spot instance requests that traditionally were needed in order to launch spot instances. The SIRs are handled under the hood by AWS so we have less things to care about.

It simplifies the logic and speeds up the launch of spot instances since we need to perform less API calls. Also the new RunInstances call accepts the tags out of the box, so we no longer need to use waiters, etc.

This should help considerably on large installations, where we previously ran into request throttling while waiting for spot instance requests to launch their instances, which rendered AutoSpotting unusable over a certain amount of groups handled in parallel.

It also seems to better handle the out of capacity errors, which currently can cause the launch of additional spot instances that fail to be added to their group and remain running, causing additional costs. 
 
This partially addresses the issues reported in #183.

   to it.
   and
   configurations.

This work is sponsored by [HERE Technologies](https://www.here.com)


cristim:
Tested by @symonds on a large setup and seems to work as expected, this appears to be much more scalable than before:
> all good this morning with RunInstance branch, 50 ASG's all started on-demand at the same time, 3 lambda runs later autospotting had replaced all 50 instances. Max Run Duration: 26s, Max Memory: 155M

I would like to get this reviewed and merged ASAP, this is significantly better than the current implementation.

cristim:
Thanks @artemnikitin for the approval.

@xlr-8 & @lenucksi do you guys have any thoughts on this? This is a large change at the very core of the project and I'd like it to be reviewed very carefully by multiple folks.

core/autoscaling_test.go
```diff
@@ -3377,61 +2982,181 @@ func TestGetDisallowedInstanceTypes(t *testing.T) {
...
-			policy:               "aggressive",
-			want:                 0.038664,
+			name: "doesn't have member",
+			Group: &autoscaling.Group{
+				Instances: []*autoscaling.Instance{
+					{InstanceId: aws.String("foo")},
+					{InstanceId: aws.String("bar")},
+					{InstanceId: aws.String("baz")},
+				},
+			},
```
xlr-8: I'd usually also try with nil/empty values - just for the sake of it.

core/instance.go
```diff
@@ -269,52 +273,222 @@ func (i *instance) getCheapestCompatibleSpotInstanceType(allowedList []string, d
...
 			debug.Println("Current best option: ", chosenSpotType, " at ", bestPrice)
 		}
 	}
 	if chosenSpotType != "" {
 		debug.Println("Cheapest compatible spot instance found: ", chosenSpotType)
-		return chosenSpotType, nil
+		return cheapest, nil
 	}
-	return chosenSpotType, fmt.Errorf("No cheaper spot instance types could be found")
+	return cheapest, fmt.Errorf("No cheaper spot instance types could be found")
```
xlr-8: The first thing I wondered is:I suppose that `cheapest` is valued to `nil` in such case. But I feel that

core/instance_test.go
```diff
@@ -1163,129 +1170,534 @@ func TestTerminate(t *testing.T) {
...
+						{
+							Key:   aws.String("launched-for-asg"),
+							Value: aws.String("mygroup"),
+						},
+					},
+				},
+				},
+			},
+		},
+	}
```
xlr-8: I feel that those tests could be enriched with all the if XXX == nil conditions - but I understand this is a pain :)

core/instance.go
xlr-8: The first thing I wondered is: "Is cheapest nil?"
I suppose that it is, but I feel that it would be better to remove the doubt, by returning explicitly nil.

Even if it isn't the code later on, check the error returned value, and doesn't use cheapest, so it is unnecessary to return it.

core/instance_test.go
cristim: Maybe next time :-)

core/instance.go
cristim: Thanks, I'll fix this soon.

PR #270, opened 2018-06-15
Set the AssociatePublicIpAddress on spot instances

### LINKED ISSUE #271: No public IP's on replaced instances
# Github issue #

## Issue type ##

- Bug Report

## Summary ##

When a instance in an ASG has a public IP the newly launched spot instance does not get a public ip

## Steps to reproduce ##

Create an ASG with instances with public ip's, wait till autospotting replace them.

## Expected results ##

Instances with public ip should have a public ip when autospotting replaces them

## Actual results ##

Instances does not get public ips

cristim:
This was missing even if configured on the on-demand instances, we lost
it when changing the code to RunInstances

# Issue Type

- Bugfix Pull Request, fixes #271 

## Summary

Set the AssociatePublicIpAddress field when launching spot instances

This was reported to be missing by @atillamas on gitter. The code snippet that was doing it was dropped by mistake when changing the code to RunInstances.

   to it.
   and
   configurations.

### DIFF
--- a/core/instance.go
+++ b/core/instance.go
@@ -409,6 +409,12 @@ func (i *instance) createRunInstancesInput(instanceType string, price float64) *
 		TagSpecifications: i.generateTagsList(),
 	}
 
+	if i.IamInstanceProfile != nil {
+		retval.IamInstanceProfile = &ec2.IamInstanceProfileSpecification{
+			Arn: i.IamInstanceProfile.Arn,
+		}
+	}
+
 	if i.asg.LaunchTemplate != nil {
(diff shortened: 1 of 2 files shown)


cristim:
@atillamas please when you have some time report a github issue about this with the required details, link it to this PR and try running the branch in this PR to see if this still fixes the problem.

atillamas:
@cristim This PR works fine, instances now get the public ip

cristim:
@atillamas thanks for the quick test, great to hear it works!

I'll merge this as soon as I get a code review approval.

PR #287, opened 2018-08-06
Amended PR for issue 138: default to TerminateInstanceInAutoScalingGroup method

otterley:
Ensure users can use legacy instance termination behavior if they
so desire, via a new command-line argument/environment variable
called `instance_termination_method`, which can be set either to
`autoscaling` (default) or `detach` (legacy behavior).

Add Terraform and CloudFormation support for the variable as well.

Add documentation for this variable.


cristim:
Merged after waiting for comments from other contributors for more than a week, but none were given.

lenucksi:
Just took a look (post-merge, sorry) and what I saw looked good and straight-forward.

PR #291, opened 2018-08-16
Check ASG min OD rules before launching new spot instance

### LINKED ISSUE #290: Seeing "stranded" spot instances when minimum OD instances are configured
# Github issue #

## Issue type ##

- Bug Report

## Build number #

Custom build based on 24a5693f0c336f32fddce4fdaab87aeccca26d4f (from Jul 26, 2018)

Diff: https://gist.github.com/jjones-smug/7fe55c7015982411a92f30b31298d529

We also use a different setup for IAM permissions, but I don't think that's relevant.  I can share those if necessary but want to avoid to much SPAM.

## Configuration ##

Our configuration is pretty simplistic/vanilla.  The one thing that I think is important for this issue is that we do specify a minimum number of OD instances for some of our ASG's.  And those are the ones where we're seeing the "stranded" spot instances.

## Environment ##

- us-east-1
- VPC

We run multiple Lambda's with different configurations, and multiple ASG's with different tags/configs as well.  And it's all in modularized Terraform configurations so it's hard to post a single launch configuration.  If it's necessary I can come up with something, but I think the issue is just related to the use of a minimum OD number for an ASG.

## Summary ##

We're seeing spot instances launched by the Autospotting lambda that are left running and not swapping into the ASG for which they were intended.  It looks like the code is launching the spot instances anytime there's at least one OD instance in the ASG.  But the subsequent code that checks to see if it's OK to swap it into place, by checking the "grace period" and the minimum number of OD instances, decides to wait indefinitely.  Today a monitoring script noticed spot instances that were running for several days without being associated with an ASG.

## Steps to reproduce ##

- Setup an ASG with min instances set to 2, max instances set to 3
- Configure the ASG with an Autospotting tag setting minimum number of OD instances to 1.
- Launch the ASG with 2 instances and wait for Autospotting to run.
- It will replace one of the instances normally with a spot instance, then launch a second spot instance.  That second spot instance will not be swapped into the ASG, since at least one OD instance must be allocated.
- Subsequent Autospotting runs will log "Waiting for next run while processing ASG" each time is scans the ASG, but leave the spot instance running.

- Then increase the ASG to 3 instances.
- Wait for the next Autospotting run and the 2nd spot instance that was launched previously will be swapped into the ASG.

## Expected results ##

I include the expected/actual results with the description.

## Actual results ##

I include the expected/actual results with the description.

I believe that one way to prevent Autospotting from launching spot instances that will be blocked from use later, is to bracket this call,

```
err := onDemandInstance.launchSpotReplacement()
```
in the process() routine in "autoscaling.go" with a call to the minimum OD instance check routine.

jjones-smug:
# Issue Type

- Bugfix Pull Request

## Summary

Fixes #290 

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -342,6 +342,12 @@ func (a *autoScalingGroup) process() {
 				"No running on-demand instances were found, nothing to do here...")
 			return
 		}
+
+		if !a.needReplaceOnDemandInstances() {
+			logger.Println("Not allowed to replace any of the running OD instances in ", a.name)
+			return
+		}
+
 		a.loadLaunchConfiguration()
 		err := onDemandInstance.launchSpotReplacement()
 		if err != nil {
--- a/core/autoscaling_test.go
+++ b/core/autoscaling_test.go
@@ -1379,6 +1379,29 @@ func TestNeedReplaceOnDemandInstances(t *testing.T) {
 			desiredCapacity: aws.Int64(4),
 			expectedRun:     true,
 		},
+		{name: "ASG has on-demand instances equal to the min on-demand number",
+			asgInstances: makeInstancesWithCatalog(
+				map[string]*instance{
+					"id-1": {
+						Instance: &ec2.Instance{
+							State:             &ec2.InstanceState{Name: aws.String(ec2.InstanceStateNameRunning)},
+							Placement:         &ec2.Placement{AvailabilityZone: aws.String("eu-west-1a")},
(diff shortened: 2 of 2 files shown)


jjones-smug:
Ah, a test case... That should be doable.

jjones-smug:
I added a test that I think covers the new condition I added to the logic.  The code added ran a clean "make full-test" when I added that test to my local repo.

PR #295, opened 2018-09-07
Support launching spot nodes of recent instance types 

### LINKED ISSUE #294: Recent instance types (T3, R5, etc.) are not supported
# Github issue #

- Bug Report

## Build number #

```text
up to build 929
```

## Configuration ##

- T3, R5, etc. instances as initial instance type set on the ASG
- other on-demand instance types that would normally be replaceable by these recent instance types

## Summary ##

AutoSpotting lacks support for the recent instance types, because we lack spec and/or pricing information for them, which should come from the upstream ec2instances.info instance type database.

## Steps to reproduce ##

- Use AutoSpotting against a group of T3 on-demand instance types with the allowed instances types configuration set to "current". 
and also
- Use AutoSpotting against a group of C5 on-demand instance types with the allowed instances types configuration set to "t3*"

## Expected results ##

For both these cases the groups should be replaced by T3 spot instances.

## Actual results ##

The groups keep running their on-demand nodes, AutoSpotting fails to launch any spot nodes for them.

cristim:
# Issue Type

- Bugfix Pull Request

## Summary

This partially fixes #294, allowing to launch spot nodes of instance types that lack on-demand pricing data. 

Previously we would skip adding the type info when missing the on-demand pricing data.

For the use case of launching spot instances we can just use the spot market prices which we already have and use the rest of the instance type information (mainly hardware specs) which we ignored so far.

   to it.
   and
   configurations.

### DIFF
--- a/core/region.go
+++ b/core/region.go
@@ -263,7 +263,7 @@ func (r *region) requestSpotPrices() error {
 		price, err := strconv.ParseFloat(*priceInfo.SpotPrice, 64)
 		if err != nil {
 			logger.Println(r.name, "Instance type ", instType,
-				"Coudln't parse spot price, may not be available on the spot market")
+				"Couldn't parse spot price, may not be available on the spot market")
 			continue
 		}
 


core/instance.go
```diff
@@ -293,8 +293,8 @@ func (i *instance) getCheapestCompatibleSpotInstanceType(allowedList []string, d
 
 	for _, candidate := range i.region.instanceTypeInformation {
 
-		logger.Println("Comparing ", candidate.instanceType, " with ",
-			current.instanceType)
+		logger.Printf("Comparing %s with %s ",
+			current.instanceType, candidate.instanceType)
```
xlr-8: Why switching this one to `Printf` but leaving the other as `Println` ?

core/region.go
```diff
@@ -206,31 +206,30 @@ func (r *region) determineInstanceTypeInformation(cfg *Config) {
...
-			info = instanceTypeInformation{
-				instanceType:        it.InstanceType,
-				vCPU:                it.VCPU,
-				memory:              it.Memory,
-				GPU:                 it.GPU,
-				pricing:             price,
-				virtualizationTypes: it.LinuxVirtualizationTypes,
-				hasEBSOptimization:  it.EBSOptimized,
-			}
+		if price.onDemand == 0 {
```
xlr-8: I'm aware it's unlikely to happen, but maybe switching to `<= 0` would be safer as even if the value is negative the code would warn us.

Also according to the previous method's behaviour, shouldn't this be followed by an `else` afterwards? Or a `return` on the `if` itself if no else is desired?

core/instance.go
cristim: I had a larger local patch in which I wanted to print on the same line the reason why the comparison failed: CPU, memory, storage, etc.

I failed to do that cleanly for now, so I reverted most of that but this hunk was missed.

core/region.go
cristim: This should never be negative, the check is for the zero value in case we're missing the data.

I'll revisit the if logic on Friday.

lenucksi: I agree with @xlr-8's question and suggestion.

gabegorelick:
Does this fix #376 (AutoSpotting ignoring spot instances that it incorrectly thinks are $0)?

cristim:
Not sure if this would help you but I can explain what's going on:

This covers a situation in which existing instance types are later added to new regions.

In this case we may have instance type information for some other regions but on demand price is zero for the current region because at the time we compiled the instance type information that particular instance type wasn't available there. If the price is zero we currently just skip the instance type in this region.

Later, once the instance becomes available we will get spot price information for it, which is all we need to launch the spot instances.

This change allows us to launch the spot instances if we have spot price information about them even if we're lacking the on demand price information

Issue #285, opened 2018-08-01
Launch Template support


symonds:
When running autospotting against an ASG which is configured using Launch Templates instead of Launch Configurations the current master 24a5693f0c336f32fddce4fdaab87aeccca26d4f branch has the following error:

```
2018/08/01 12:36:28 instance.go:331: Launching spot instance with a bid = 0.1
2018/08/01 12:36:28 instance.go:312: Couldn't launch spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: 786d65a8-b6c1-4489-ad73-42ee66ef268f
2018/08/01 12:36:28 autoscaling.go:337: Could not launch cheapest spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: 786d65a8-b6c1-4489-ad73-42ee66ef268f
2018/08/01 12:36:28 instance.go:312: Couldn't launch spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: dd109c9c-4a26-4535-ae1d-b1377a7b3ce7
2018/08/01 12:36:28 autoscaling.go:337: Could not launch cheapest spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: dd109c9c-4a26-4535-ae1d-b1377a7b3ce7
2018/08/01 12:36:28 instance.go:312: Couldn't launch spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: b3ab266e-839c-4f05-aa48-a858d8693ae6
2018/08/01 12:36:28 autoscaling.go:337: Could not launch cheapest spot instance: InvalidParameterCombination: A launch template ID and a launch template name cannot be specified in the same request.
status code: 400, request id: b3ab266e-839c-4f05-aa48-a858d8693ae6
2018/08/01 12:36:28 Execution completed, nothing left to do
END RequestId: 81d0857d-9587-11e8-af34-2b4da098a743
```

Seems to be related to https://github.com/cristim/autospotting/blob/master/core/instance.go#L422

cristim:
Good catch, thanks for testing this. 

I wasn't aware that just one of those is required and I have no environments based on launch templates where I could test this.

Try to delete the line with the name, and please create a PR if it works as expected.

[...]

PR #303, opened 2018-10-19
Add support for environments using Launch Templates

xbglowx:
* Fixes https://github.com/cristim/autospotting/issues/285
* Add appropriate EC2 tag depending on if we are using launch config or
  template

# Issue Type

- Bugfix Pull Request

## Summary

Fixes https://github.com/cristim/autospotting/issues/285, which would prevent usage of launch templates, due to specifying both the ID and name at the same time.

   to it.
   and
   configurations.

### DIFF
--- a/core/instance.go
+++ b/core/instance.go
@@ -534,16 +534,13 @@ func (i *instance) createRunInstancesInput(instanceType string, price float64) *
 		TagSpecifications: i.generateTagsList(),
 	}
 
-	if i.IamInstanceProfile != nil {
-		retval.IamInstanceProfile = &ec2.IamInstanceProfileSpecification{
-			Arn: i.IamInstanceProfile.Arn,
-		}
-	}
-
 	if i.asg.LaunchTemplate != nil {
 		retval.LaunchTemplate = &ec2.LaunchTemplateSpecification{
-			LaunchTemplateId:   i.asg.LaunchTemplate.LaunchTemplateId,
-			LaunchTemplateName: i.asg.LaunchTemplate.LaunchTemplateName,
+			LaunchTemplateId: i.asg.LaunchTemplate.LaunchTemplateId,
+		}
+	} else if i.IamInstanceProfile != nil {
+		retval.IamInstanceProfile = &ec2.IamInstanceProfileSpecification{
+			Arn: i.IamInstanceProfile.Arn,
 		}
 	}
 
@@ -587,10 +584,6 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
 		ResourceType: aws.String("instance"),
 		Tags: []*ec2.Tag{
 			{
-				Key:   aws.String("LaunchConfigurationName"),
-				Value: i.asg.LaunchConfigurationName,
-			},
-			{
 				Key:   aws.String("launched-by-autospotting"),
 				Value: aws.String("true"),
 			},
@@ -599,6 +592,18 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
 				Value: aws.String(i.asg.name),
 			},
 		},
+	}
+
+	if i.asg.LaunchTemplate != nil {
+		tags.Tags = append(tags.Tags, &ec2.Tag{
+			Key:   aws.String("LaunchTemplateName"),
+			Value: i.asg.LaunchTemplate.LaunchTemplateName,
+		})
+	} else if i.asg.LaunchConfigurationName != nil {
+		tags.Tags = append(tags.Tags, &ec2.Tag{
+			Key:   aws.String("LaunchConfigurationName"),
+			Value: i.asg.LaunchConfigurationName,
+		})
 	}
 
 	for _, tag := range i.Tags {
--- a/core/instance.go
+++ b/core/instance.go
@@ -537,8 +537,11 @@ func (i *instance) createRunInstancesInput(instanceType string, price float64) *
 	if i.asg.LaunchTemplate != nil {
 		retval.LaunchTemplate = &ec2.LaunchTemplateSpecification{
 			LaunchTemplateId: i.asg.LaunchTemplate.LaunchTemplateId,
+			Version:          i.asg.LaunchTemplate.Version,
 		}
-	} else if i.IamInstanceProfile != nil {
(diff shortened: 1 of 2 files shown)


core/instance.go
```diff
@@ -504,6 +497,20 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+			Value: i.asg.LaunchTemplate.LaunchTemplateName,
+		})
+	}
+
+	if i.asg.LaunchConfigurationName != nil {
+		tags.Tags = append(tags.Tags, &ec2.Tag{
+			Key:   aws.String("LaunchConfigurationName"),
+			Value: i.asg.LaunchConfigurationName,
+		})
+	}
```
xlr-8: Logic wise wouldn't that be an `if / else if` - rather than 2 if?

xlr-8: Logic wise wouldn't that be an `if / else if` - rather than 2 if?

xbglowx: AWS will make sure an ASG can only have one or the other, so the if/else or if/if will act the same, from what I have seen. That being said, if you prefer if/else I can fix the PR?

xbglowx: AWS will make sure an ASG can only have one or the other, so the if/else or if/if will act the same, from what I have seen. That being said, if you prefer if/else I can fix the PR?

xlr-8: I mean it's nothing big, but as the PR will be re-updated for adding tests, then I guess it could be updated.

As you said:
> AWS will make sure an ASG can only have one or the other [...]

It's indeed purely from a logical point of view, if it can only be one of them then better add an `if / else if` to make it explicit, rather than 2 if which somehow imply that both could happen.

xbglowx: I agree.

xbglowx:
I will try to fix/update tests soon.

xbglowx:
@cristim I believe I ran into a problem with IamInstanceProfile not being set when I tried using launch template, so I assumed they were. Unfortunately, I am on paternity leave and won't be able to reproduced or work on this PR until I am back at work.

Sorry for leaving this PR unfinished.

bi0morph:
Hey guys,

I'm just curious - Do you plan review/merge this PR?
😼

core/instance.go
```diff
@@ -462,16 +462,13 @@ func (i *instance) createRunInstancesInput(instanceType string, price float64) *
...
-		retval.IamInstanceProfile = &ec2.IamInstanceProfileSpecification{
-			Arn: i.IamInstanceProfile.Arn,
-		}
-	}
-
 	if i.asg.LaunchTemplate != nil {
 		retval.LaunchTemplate = &ec2.LaunchTemplateSpecification{
-			LaunchTemplateId:   i.asg.LaunchTemplate.LaunchTemplateId,
-			LaunchTemplateName: i.asg.LaunchTemplate.LaunchTemplateName,
+			LaunchTemplateId: i.asg.LaunchTemplate.LaunchTemplateId,
```
dpikedpike: There is a subtle bug here. getting the launch template id from the LaunchTemplateSpecification without specifying the version means you will get the Default version of the launch template, which may be *far* different that the version currently being used.

cristim: @dpikedpike good catch!

I'm not very familiar with the launch templates yet, can you suggest how we could improve this to get the currently used launch template version?

Feel free to add to patches addressing this issue on top of this PR.

cristim:
@bi0morph if you're interested feel free to pick it up, the tests are currently failing and need some attention.

bi0morph:
> @bi0morph if you're interested feel free to pick it up, the tests are currently failing and need some attention.

I'll try to review and suggest something. I'm not good at `Go` but I tried and launched `LaunchTemplateId` a few weeks ago.

core/instance.go
dpikedpike: Code should be:
		retval.LaunchTemplate = &ec2.LaunchTemplateSpecification{
			LaunchTemplateId: i.asg.LaunchTemplate.LaunchTemplateId,
			Version:          i.asg.LaunchTemplate.Version,
		}

The version then comes from the version of the LT being used by the asg

cristim:
This PR is getting old, is anyone passionate about this enough to pick it up and get it merged?

bi0morph:
@cristim I will try to do something with that in two weeks.

core/instance.go
cristim: Fixed in my latest commit

core/instance.go
```diff
@@ -537,8 +537,11 @@ func (i *instance) createRunInstancesInput(instanceType string, price float64) *
 	if i.asg.LaunchTemplate != nil {
 		retval.LaunchTemplate = &ec2.LaunchTemplateSpecification{
 			LaunchTemplateId: i.asg.LaunchTemplate.LaunchTemplateId,
+			Version:          i.asg.LaunchTemplate.Version,
 		}
-	} else if i.IamInstanceProfile != nil {
```
cristim: I think this was a bug, I doubt Launch Templates and the IAM Instance Profile should be mutually exclusive.

cristim:
I had a few spare cycles to spend on this and I did a bit of progress, but I didn't test this yet.

@bi0morph, @dpikedpike, @xbglowx please let me know what you think, and if you could test this it would be great.

tathagata:
I have a couple of stacks to test with LaunchTemplates. I built this pr version locally.
I was initially getting 
```
No spot instances were found for <scaling-group-name>
```
and then on changing min in my scaling group to 1, spot instances were being requested. Had to remove scale in protection - no other issues.
I think when I tried it the last time in October 2018, I got the error message that's described in original bug description - but did not face that issue now.
Any specific testing that would help?

cristim:
Thanks @tathagata, that sounds great. Everything you described is just like expected

tathagata:
OT:
The `coverage/coveralls` check [seems to be something](https://github.com/lemurheavy/coveralls-public/issues/1183#issuecomment-420158040) to do with the Oauth settings. May be moving to the repo to AutoSpotting organization needed some additional config changes?

cristim:
I think the coveralls status is only available for the master branch because of the way TravisCI protects secret values.

Issue #263, opened 2018-06-07
RunInstances API call, attempt multiple instance types


symonds:
## Issue type ##

- Feature Idea

## Summary ##

When using the new RunInstances API call (https://github.com/cristim/autospotting/pull/257) and there is high demand on the instance type AWS responds with `InsufficientInstanceCapacity`

Which Autospotting returns

```
2018/06/07 11:56:39 instance.go:306: Couldn't launch spot instance: InsufficientInstanceCapacity: There is no Spot capacity available that matches your request.
	status code: 500, request id: 3045e3ba-c3f0-40d1-903b-92327ec3b5c1
2018/06/07 11:56:39 autoscaling.go:337: Could not launch cheapest spot instance: InsufficientInstanceCapacity: There is no Spot capacity available that matches your request.
	status code: 500, request id: 3045e3ba-c3f0-40d1-903b-92327ec3b5c1
```

It would be nice if Autospotting went back to `getCheapestCompatibleSpotInstanceType` with the failed instance type appended to `disallowedList` and attempted to bid for the next cheapest instance.

I would limit this action to 3 attempts.

cristim:
This is a great proposal, thanks!

cristim:
I'm thinking to just iterate over the list of compatible instance types ordered increasingly by price and launch them one by one until we get one that can be launched successfully.

sunnynew:
Hi @cristim 

First of all congrats on this wonderful tool. 

I wanted to know about this particular issue, anytime soon can we expect a fix.  As you mentioned just make Autospotting to iterate over all instance type mentioned in allowed_instance_types list. 

This will be a lifesaver. 

Thanks!

cristim:
@sunnynew thanks for your feedback, I appreciate it.

I'm now focusing on #294 so I'm not planning to work on this in the very near future, but feel free to take it over and contribute a PR if you need this so badly and have some time to spare.

I can help you with guidance if you need any.

cristim:
Fixed in #306

PR #306, opened 2018-11-06
Attempt multiple instance types for launch. #263

lenucksi:
# Issue Type
- Feature Pull Request

## Summary

This provides and implementation approach to the problem described in #263. 
It's the first form passing tests clean and probably has enough non idiomatic Go in it. I blame Haskell for that. ;)

This PR is my relevant commit cherrypicked from my other branch onto the last master. It passes ```go fmt``` clean and should generally be cleaner. 

1. [X] The contribution fixes a single existing github issue, and it is linked
   to it.
1. [X] No issues are reported when running `make full-test`.


core/instance.go
```diff
@@ -291,36 +296,46 @@ func (i *instance) getCheapestCompatibleSpotInstanceType(allowedList []string, d
...
+			result = append(result, ai.instanceTI)
+		}
+		return result, nil
 	}
-	return cheapest, fmt.Errorf("No cheaper spot instance types could be found")
+
+	return nil, fmt.Errorf("No cheaper spot instance types could be found")
 }
 
 func (i *instance) launchSpotReplacement() error {
```
cristim: The test coverage decreased because this function is not covered and it's now a bit bigger than before. 

I'd like to have this covered as well so that we're not decreasing our test coverage.

PR #309, opened 2018-12-20
Detach instance when spot termination notification is received

adeshrd:
When a spot instance termination event is triggered, the lambda will get the
region and the instance id from the event data. With this data, it will detach
that instance.

# Issue Type

- Feature Pull Request

## Summary

The template now contains another CloudWatch EventRule to trigger the lambda 
when a spot instance termination notification is received.

The lambda detaches the instance when it receives the termination notification from
CloudWatch.

Note: The new EventRule is only present in the cloudformation template. 

   to it.
   and
   configurations.

### DIFF
--- a/autospotting.go
+++ b/autospotting.go
@@ -99,8 +99,8 @@ func Handler(ctx context.Context, event events.CloudWatchEvent) {
 	}
 
 	if instanceID != nil {
-		spotTermination := autospotting.SpotTermination{}
-		spotTermination.DetachInstance(instanceID, event.Region)
+		spotTermination := autospotting.NewSpotTermination(event.Region)
+		spotTermination.DetachInstance(instanceID)
 	} else {
 		run()
 	}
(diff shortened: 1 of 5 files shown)


adeshrd:
Thanks please do. Since this is my first time on Go I might have missed some things. Till then I will work on getting that coverage done.

cristim:
The code looks good to me, I'm just not sure how to handle the Cloudwatch event situation I mentioned on gitter.

The spot termination event needs to be created in all regions and linked to the Autospotting Lambda function deployed in a single region. We have to find a way to deploy it, maybe using a custom Cloudformation resource. I am also not sure how to handle this on terraform

adeshrd:
This was more troublesome than I thought but I was able to create those EventRules across regions. Have mentioned in more detail in commit comments

cristim:
I decided to merge this as it is, if anyone wants the old behavior badly enough to submit a PR to make it configurable, I'd happily consider it.

PR #316, opened 2019-02-06
Use dynamic names for terraform resources so the module can be reused

### LINKED ISSUE #315: Feature: multiple lambda instances (with different configurations)
## Use case ##

Running multiple ASGs with different instance type requirements.

## Problem ##

In the terraform module (at least), several resource names are hard-coded (e.g., function name, role name, policy name) so the module cannot be easily used with multiple configurations.

## Proposal ##

Use variables so the module can be reused with different configurations.

bfin:
Fixes #315

### DIFF
--- a/terraform/label.tf
+++ b/terraform/label.tf
@@ -1,5 +1,5 @@
 module "label" {
-  source      = "git::https://github.com/cloudposse/terraform-null-label.git?ref=master"
+  source      = "git::https://github.com/cloudposse/terraform-null-label.git?ref=0.5.4"
   context     = "${var.label_context}"
   namespace   = "${var.label_namespace}"
   environment = "${var.label_environment}"
--- a/terraform/label.tf
+++ b/terraform/label.tf
@@ -1,0 +1,11 @@
+module "label" {
+  source      = "git::https://github.com/cloudposse/terraform-null-label.git?ref=master"
+  context     = "${var.label_context}"
+  namespace   = "${var.label_namespace}"
+  environment = "${var.label_environment}"
+  stage       = "${var.label_stage}"
+  name        = "${var.label_name}"
+  attributes  = ["${var.label_attributes}"]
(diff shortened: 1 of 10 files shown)


bfin:
Per the discussion in #315, there's a redundant terraform module in this repo (one is in the `terraform` directory, then another in the `terraform/autospotting` subdirectory). In anticipation of refactoring, I defined some label variables once in the top-level module (`terraform`) and then used the context feature of the `null-label` module in the submodule (`terraform/autospotting`). This provides somewhat different functionality for the 2 modules: the top-level module can receive the namespace, environment, name, etc. directly, while the submodule is only passed the context (which includes all of the above labels). This made it so the variables didn't have to be defined twice, but prevents the submodule from receiving the default label values when used directly (as referenced in the docs). I'll re-define the variables in the submodule for now so it's backwards-compatible and update this PR shortly.

bfin:
Ok, now we have the same functionality between the module and submodule, and it's backwards-compatible. By default, the function/role/etc. name will be still be `autospotting` but this can be configured by passing any of the following variables:
- `label_context`
- `label_namespace`
- `label_environment`
- `label_stage`
- `label_name`
- `label_attributes`
- `label_tags`
- `label_delimiter`

terraform/label.tf
```diff
@@ -0,0 +1,11 @@
+module "label" {
+  source      = "git::https://github.com/cloudposse/terraform-null-label.git?ref=master"
```
cristim: I'd like us to vendor this by pinning to a given release/commit for this dependency, in order to avoid upstream breakages.

bfin: I don't think they've ever released a backwards-incompatible change, but you're probably right to be safer here. Will do.

bfin: Done

cristim:
@bfin thank you for your contribution to the project!

PR #325, opened 2019-02-15
Terminate unused spot instances #321

### LINKED ISSUE #321: Terminate unused spot instances
# Github issue #

## Issue type ##

- Feature Idea

## Build number #

master

## Environment ##

- AWS region
- Type of environment: (VPC, EC2Classic or DefaultVPC)

```text
```

## Summary ##

(I have not tested it, but looking at code and using some logic behavior should be similar to the one described here)

It's partially related to the situation of issue #320 

Let's assume our ASG is configured for 3 desired min 2 max 4, and an autoscaling policy (tracking or step) is configured.

1) Autospot launch a spot instance
2) Before the next AutoSpot invocation, Autoscaling policy reduce the number of instances and terminate the ondemand one (the one that the next invocation of autospot should replace).
3) Autospot do not see more ondemand instance to replace and spot instance launched in point 1 remain active and unused.

i think the "logic" is in core/autoscaling.go func process lines 360:

        if !a.needReplaceOnDemandInstances() || !spotInstance.isReadyToAttach(a) {
                logger.Println("Waiting for next run while processing", a.name)
                return
        }

We should separate the or condition and process the two cases differently:
```
if !spotInstance.isReadyToAttach(a) {
    logger.Println("Waiting for next run while processing", a.name)
    return
}
if !a.needReplaceOnDemandInstances() {
    "here we shoud terminate the unused instance"
}
```

Maybe you already thought about it but there was some problems that currently i am missing...
In that case please le me know.

Regards, Alberto.

mello7tre:
# Issue Type

- Feature Pull Request

## Summary

Currently.
A spotted instance is launched.
Before the next autospot execution the ondemand instance that autospot
next execution should have replaced is terminated.
The launched spot instance, that is no more necessary, is not terminated.

The idea is:
once we found a spot instance launched for an ASG, we first check if the
ASG still need it.
If not we terminate the spot instance.
Beacuse we cannot know if the ASG will need it in the future and after
how much time.

To test:
create an ASG with min=max=desired=1
wait until autospot launch a spot for that ASG
set min=desired=0
look at next autospot execution log for a line like:
```
2019/02/15 13:33:24 autoscaling.go:366: Spot instance i-0fa844e5c281023a0 is not need anymore by ASG ecs-s-d-AutoScalingGroup-DLZ9BYSWB9V4 terminating the spot instance.
```
check that launched spot instance is terminated.

fix #321

1. [X ] The contribution fixes a single existing github issue, and it is linked
   to it.
1. [ X] The code is as simple as possible, readable and follows the idiomatic Go
1. [ X] No issues are reported when running `make full-test`.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -357,7 +357,13 @@ func (a *autoScalingGroup) process() {
 
 	spotInstanceID = *spotInstance.InstanceId
 
-	if !a.needReplaceOnDemandInstances() || !spotInstance.isReadyToAttach(a) {
+	if !a.needReplaceOnDemandInstances() {
+		logger.Println("Spot instance", spotInstanceID, "is not need anymore by ASG",
+			a.name, "terminating the spot instance.")
+		spotInstance.terminate()
+		return
+	}
+	if !spotInstance.isReadyToAttach(a) {
 		logger.Println("Waiting for next run while processing", a.name)
 		return
 	}
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -357,14 +357,14 @@ func (a *autoScalingGroup) process() {
 
 	spotInstanceID = *spotInstance.InstanceId
 
-	if !a.needReplaceOnDemandInstances() {
-		logger.Println("Spot instance", spotInstanceID, "is not need anymore by ASG",
-			a.name, "terminating the spot instance.")
-		spotInstance.terminate()
-		return
-	}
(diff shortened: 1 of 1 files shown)


cristim:
I think the code looked cleaner before the Code climate fix. I  could have approved it with the additional return statement. Feel free to revert if you agree.

mello7tre:
Ok thanks, i liked it better too.

PR #328, opened 2019-02-21
Termination Notification Configurable Action

mello7tre:
# Issue Type

- Feature Pull Request

## Summary

Idea:
At the moment, when a termination notification message is received,
we are only supporting detach as action on the terminating spot instance.

Detaching an instance do not trigger Lifecycle Hook (LH).

This patch add a new option "termination_notification_action" with
relative CloudFormation parameter "TerminationNotificationAction" that
support the options:
```
detach
	current behaviour - no LH Triggered
terminate
	new behaviour - spot instance is terminated without decrementing
	desired capacity, so that ASG spawn a new instance (same as detach) - LH
	Triggered.
auto
	new behaviour - if ASG has a LH for Terminate Lifecycle
	Transition, action is "terminate", if not action is detach.
```
The option is global for all ASG.
But using "auto", behaviour is based on the presence of LH, this way it
can act in different ways.

   to it.
   and
   configurations.

### DIFF
--- a/cloudformation/stacks/AutoSpotting/template.yaml
+++ b/cloudformation/stacks/AutoSpotting/template.yaml
@@ -40,6 +40,17 @@
         "Instance termination method.  Must be one of 'autoscaling' (default) or
         'detach' (compatibility mode, not recommended)"
       Type: "String"
+    TerminationNotificationAction:
+      AllowedValues:
+        - "auto"
+        - "detach"
+        - "terminate"
+      Default: "terminate"
+      Description: >
+        "Action to do when receiving a Spot Instance Termination Notification.
+        Must be one of 'auto' (terminate if lifecyclehook else detach) [default],
+        'terminate' (lifecyclehook triggered), 'detach' (lifecyclehook not triggered)"
+      Type: "String"
     FilterByTags:
       Default: ""
       Description: >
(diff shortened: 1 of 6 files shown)


cloudformation/stacks/AutoSpotting/template.yaml
```diff
@@ -40,6 +40,17 @@
         "Instance termination method.  Must be one of 'autoscaling' (default) or
         'detach' (compatibility mode, not recommended)"
       Type: "String"
+    TerminationNotificationAction:
+      AllowedValues:
+        - "auto"
+        - "detach"
+        - "terminate"
+      Default: "terminate"
```
cristim: I would set the default to "auto"

core/spot_termination.go
```diff
@@ -101,3 +116,57 @@ func (s *SpotTermination) getAsgName(instanceID *string) (string, error) {
...
+			s.terminateInstance(instanceID, asgName)
+		} else {
+			s.detachInstance(instanceID, asgName)
+		}
+	}
+
+	return nil
+}
+
+func (s *SpotTermination) asgHasHook(autoScalingGroupName *string) bool {
```
cristim: I would rename this to `asgHasTerminationLifecycleHook` or if we keep the name as it is, we can make it take the hook type as an additional parameter.

cloudformation/stacks/AutoSpotting/template.yaml
mello7tre: Sure, that is the right one!
Just forgot to make the change.

core/spot_termination.go
mello7tre: Ok. As the only LF that make sense, in that contest, is the Termination one i will rename it as your suggestion.

PR #329, opened 2019-02-27
Delete Tag 'launched-for-asg' when detaching a spot instance

### LINKED ISSUE #327: Terminating detached spot instance used to replace on demand one
# Github issue #

## Issue type ##

- Bug Report

## Build number #

master

## Environment ##

- AWS region
- Type of environment: (VPC, EC2Classic or DefaultVPC)

```text
```

## Summary ##

Following a spot termination notification, a previously launched spot instance has been detached from ASG, without decrementing ASG desired capacity so that ASG launch a new instance.

Happen that the new instance is launched in the same AZ that the spot terminating one.
Happen that before detached terminating spot instance terminate (so before 2 minutes), AutoSpotting is executed by scheduled event.

AutoSpotting (AS) look for spot instance ready do attach.
AS find the terminating spot instance:
it have the right tag "launched-for-asg" that match the ASG name and is state is still running.
AS find the ondeman one with the same AZ.
AS replace the ondemand instance with a nearly termination one!!!

AS and ASG log/events.
```
AS 2019-02-20T07:01:54.789Z 2019/02/20 07:01:54 Detached instance i-02a941d2d30a40a3c successfully
ASG At 2019-02-20T07:01:54Z instance i-02a941d2d30a40a3c was detached in response to a user request.
ASG At 2019-02-20T07:02:19Z Launching a new EC2 instance: i-095878a259a6c9e2b - an instance was started in response to a difference between desired and actual capacity, increasing the capacity from 0 to 1.
AS 2019/02/20 07:02:32 autoscaling.go:442: rmq-a-d-AutoScalingGroup-1HSLJRZ51V8Q7 found on-demand instance i-095878a259a6c9e2b replacing with new spot instance i-02a941d2d30a40a3c
```

## Steps to reproduce ##

Create ASG with desired=min=max=1
Wait until AS replace the ondemand with a spot one.
One minute before next AS execution, pubblish the following message on sns spot termination notification Topic (replacing spot-instance-id with the right one):
```
{
  "detail": {
    "instance-id": "spot-instance-id",
    "instance-action": "terminate"
  }
}

```
Look for AS log and ASG events.
If the new ondemand instance is launched in the same AZ, the events previously described should happen. 

## Expected results ##

AutoSpotting should not use a terminating previously launched spot instance to replace an on demand one.

## Actual results ##

AutoSpotting have used a terminating previously launched spot instance to replace an on demand one.

## Possible Fix ##
I am working to add the option to choose between detaching or terminating a spot instance when a spot notification message is received.
Terminating it had the advantage of triggering lifecyclehook and to avoid this problem.
But if we want to keep detach as possible action we need to fix this.

One way should be to remove the tag "launched-for-asg" after attaching a spot instance successfully to an ASG.

mello7tre:
# Issue Type

- Bugfix Pull Request

## Summary

In some situation (described in detail in #327), a detaching near termination
spot instance can be used to replace an ondemand one.

This fix delete the Tag 'launched-for-asg' after successfully detaching
a spot instance.

Fixes #327

1. [X] The contribution fixes a single existing github issue, and it is linked
   to it.
1. [X] All new functionality is covered by automated test cases so the overall
   and
   configurations.

### DIFF
--- a/core/spot_termination.go
+++ b/core/spot_termination.go
@@ -172,7 +172,7 @@ func (s *SpotTermination) deleteTagInstanceLaunchedForAsg(instanceID *string) er
 		return err
 	}
 
-	log.Printf("Tag 'launched-for-asg' delete from spot instance %s", *instanceID)
+	log.Printf("Tag 'launched-for-asg' deleted from spot instance %s", *instanceID)
 
 	return nil
 }
--- a/core/spot_termination.go
+++ b/core/spot_termination.go
@@ -10,6 +10,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 )
 
 const (
@@ -20,7 +22,8 @@ const (
 
 //SpotTermination is used to detach an instance, used when a spot instance is due for termination
 type SpotTermination struct {
-	asSvc autoscalingiface.AutoScalingAPI
+	asSvc  autoscalingiface.AutoScalingAPI
+	ec2Svc ec2iface.EC2API
 }
 
 //InstanceData represents JSON structure of the Detail property of CloudWatch event when a spot instance is terminated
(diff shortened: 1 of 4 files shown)


core/spot_termination.go
```diff
@@ -147,6 +154,29 @@ func (s *SpotTermination) ExecuteAction(instanceID *string, terminationNotificat
...
+		},
+	}
+	_, err := s.ec2Svc.DeleteTags(&ec2Params)
+
+	if err != nil {
+		logger.Printf("Failed to delete Tag 'launched-for-asg' from spot instance %s with err: %s\n", *instanceID, err.Error())
+		return err
+	}
+
+	log.Printf("Tag 'launched-for-asg' delete from spot instance %s", *instanceID)
```
cristim: I would say delete**d** here

PR #336, opened 2019-03-30
Updated the doco to package default or custom binaries for a lambda function

vikas027:
Updated the doco to package default or custom binaries for a lambda function
This is just for advanced users, but still, it is good to have it for clarity

# Issue Type
- Documentation Pull Request

### DIFF
--- a/CUSTOM_BUILDS.md
+++ b/CUSTOM_BUILDS.md
@@ -88,11 +88,14 @@ your CloudFormation or Terraform stack to use those new binaries.
 1. The build system can use a `BUCKET_NAME` variable that tells it where to
    upload new binaries. Set it into your environment to the name of your S3
    bucket.
-
    `export BUCKET_NAME=my-bucket`
 
 1. Define some AWS credentials or profile information into your
    [environment](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment).
+
+1. (Optional) You can also create the lambda deployment package using the below
+   command, the zip file is generated in the `build` directory.
+   `make archive`
 
 1. Build and upload your binaries to the S3 bucket.
    `make upload`


CUSTOM_BUILDS.md
```diff
@@ -83,6 +83,8 @@ your CloudFormation or Terraform stack to use those new binaries.
 
 ## Using your own binaries in AWS Lambda ##
 
+You can either follow the below process to upload the lambda package to a S3 bucket, or compile the binaries as shown [above](#compiling-the-binaries-locally) and create a lambda deployment package by zipping (`zip autospotting.zip autospotting`) the built binary.
```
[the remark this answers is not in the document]
vikas027: I have fixed this now.

CUSTOM_BUILDS.md
```diff
@@ -83,6 +83,11 @@ your CloudFormation or Terraform stack to use those new binaries.
 
 ## Using your own binaries in AWS Lambda ##
 
+You can either follow the below process to upload the lambda package to a
+S3 bucket, or compile the binaries as shown [above](#compiling-the-binaries-locally)
+and create a lambda deployment package by zipping (`zip autospotting.zip autospotting`)
```
cristim: We have the `make archive` target that already creates the zip file, we can just use that.

The file can then be found under the `build` directory.

The `BUCKET_NAME=my-bucket make upload` target will build and then automatically upload it to your own S3 bucket.

vikas027: Ah cool, I didn't know that. Let me add that to the doco.

vikas027:
Sure, that makes sense but what if someone does not wants to use an S3 bucket. I think it does not harm to keep that optional step, just in case someone wants some flexibility.

PR #338, opened 2019-04-17
Template simplifications on Fix/termination signal

mello7tre:
* regional_template.yaml/template.yaml.
  As role are cross region, there is no need to create a LambdaExecutionRole
  and relative LambdaPolicy in all regions. We can just create one in the main
  template's region and pass it's arn as parameter to the lambda that create the regional ones.

* regional_template.yaml.
  Now that we have a lambda in the same region of the event,
  there is no need to use sns at all, regional CW events can send directly
  to regional lambda.

  To avoid change in autospotting.go's code, CW event is "encapsulated"
  as an SNS message.

* template.yaml/regional_stack_lambda.py.
  To test all this i needed to add a new template parameter splitting
  the LambdaZipPath one in LambdaS3BucketPrefix and LambdaZipName.

  LambdaS3BucketPrefix is the bucket prefix from which are loaded all
  necessary resources (regional lambda zip and regional template).
  So the path, name apart, is no more hardcoded in the code/template.

During development i had some problems executing cfn.delete_stack (in
regional_stack_lambda.py) so i have choose to insert it in a try block.

# Issue Type

- Feature Pull Request

   to it.
   and
   configurations.

### DIFF
--- a/cloudformation/stacks/AutoSpotting/regional_stack_lambda.py
+++ b/cloudformation/stacks/AutoSpotting/regional_stack_lambda.py
@@ -72,13 +72,26 @@ def handle_create(event):
     role_arn = event['ResourceProperties']['LambdaRegionalExecutionRoleARN']
     bucket = event['ResourceProperties']['S3Bucket']
     path = event['ResourceProperties']['S3BucketPrefix']
-    template_url = 'https://' + bucket + '.s3.amazonaws.com/' + path + '/regional_template.yaml'
+    template_url = (
+        'https://' +
+        bucket +
+        '.s3.amazonaws.com/' +
+        path +
+        '/regional_template.yaml'
+    )
     threads = []
 
     # create concurrently in all regions
     for region in ec2.describe_regions()['Regions']:
-        process = Thread(target=create_stack,
-                         args=[region['RegionName'], lambda_arn, role_arn, template_url])
+        process = Thread(
+            target=create_stack,
+            args=[
+                region['RegionName'],
+                lambda_arn,
+                role_arn,
+                template_url,
+            ]
+        )
         process.start()
         threads.append(process)
 
--- a/cloudformation/stacks/AutoSpotting/regional_stack_lambda.py
+++ b/cloudformation/stacks/AutoSpotting/regional_stack_lambda.py
@@ -17,11 +17,9 @@ from botocore.vendored.requests.exceptions import RequestException
 SUCCESS = "SUCCESS"
 FAILED = "FAILED"
 STACK_NAME = 'AutoSpottingRegionalResources'
-TEMPLATE_URL = \
-    'https://s3.amazonaws.com/cloudprowess/nightly/regional_template.yaml'
 
 
-def create_stack(region, lambda_arn):
+def create_stack(region, lambda_arn, role_arn, template_url):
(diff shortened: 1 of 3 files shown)


mello7tre:
Hi @cristim, we you have some time can you please have a look at why travisCI is failing ?
Looking at logs seems a problem with coveralls, not related to my PR but more general:
```
autospotting_1  | HTML output written to /tmp/cover653316840/coverage.html
autospotting_1  | Bad response status from coveralls: 422
autospotting_1  | {"message":"Couldn't find a repository matching this job.","error":true}
```
search on the web show that could be related to coveralls secret token.

Thanks.

mello7tre:
Yes i have tested it.
I have created a new stack.
Checked that both regional stack and main one where created successfully, for regional looked in a region different for the main one too.
Checked that all lambdas permission where right.
Submitted a sample sns message to the regional lambda, one in the same region of the main lambda and one in a different region.

All worked as expected. Main lambda received the message and begin terminating/detaching the terminating spot instances present in the sample sns message.

(I have ready a PR for better writing method Handler in autospotting.go, so that is no more needed to encapsulate the event in an sns message, but i prefer to submit it as a separate PR after this one got merged.)

cristim:
> Hi @cristim, when you have some time can you please have a look at why travisCI is failing ?
> Looking at logs seems a problem with coveralls, not related to my PR but more general:
> 
> ```
> autospotting_1  | HTML output written to /tmp/cover653316840/coverage.html
> autospotting_1  | Bad response status from coveralls: 422
> autospotting_1  | {"message":"Couldn't find a repository matching this job.","error":true}
> ```
> search on the web show that could be related to coveralls secret token.
> 
> Thanks.
This is strange, but it somehow works after the code was merged and in case I create the PR.

It may be a way to ensure that the secret value (in our case the one storing the coveralls token) is not easy to leak by anyone who creates a PR.

cristim:
See here: https://docs.travis-ci.com/user/encryption-keys

"Please note that encrypted environment variables are not available for pull requests from forks."

mello7tre:
understand, so from now on for "external PR" travisCI will always fail ?

cristim:
That seems to be the case, but we can always change the build scripts to only run this coveralls step if the required variable is defined, having code coverage published for master should be enough, that's all we use in the github coverage badge.

cristim:
@mello7tre please create a PR with the below patch.
This should fix it, but unfortunately I can't test it since that variable would be defined if I create the PR

```
diff --git a/Makefile b/Makefile
index 715b60d..db7a27d 100644
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,9 @@ html-cover: test                                             ## Display coverage
 .PHONY: html-cover
 
 travisci-cover: html-cover                                   ## Test & generate coverage in the TravisCI format, fails unless executed from TravisCI
+ifdef COVERALLS_TOKEN
        @goveralls -coverprofile=$(COVER_PROFILE) -service=travis-ci -repotoken=$(COVERALLS_TOKEN)
+endif
 .PHONY: travisci-cover
 
 travisci-checks: fmt-check vet-check lint                    ## Pass fmt / vet & lint format
```

mello7tre:
I created it, seems to have worked.

Issue #344, opened 2019-06-04
Beanstalk initialization issue after CloudFormation update on 5/30/2019


jawadst:
## Issue type ##

- Bug Report

## Build number #

build nightly-1038

## Environment ##

eu-central-1c with Beanstalk / ECS / EC2

## Summary ##

Autospotting used to work perfectly with Beanstalk and I was a happy man.
The Beanstalk setup is the following: it uses CloudFormation to create an auto-scaling group. When an EC2 instance is launched by the auto-scaling group, it uses `cfn-init` (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-init.html) to configure itself as part of its UserData script (https://s3.dualstack.eu-central-1.amazonaws.com/elasticbeanstalk-env-resources-eu-central-1/stalks/eb_docker_ecs_4.2.1.200714.0/lib/UserDataScript.sh).
AutoSpotting applied to the auto-scaling group created by Beanstalk/CloudFormation used to work perfectly.

On 5/30/2019, AWS CloudFormation changed the way it authorizes requests from `cfn-init` and now requires the instance calling it to be a part of the CloudFormation stack it is trying to reference. This means that the instance has to have a `aws:cloudformation:stack-id` tag and that CloudFormation only allows calls from instances tagged that way (that is known as the `CFN signer` and is a special, internal-only, undocumented feature of CloudFormation for authorizing API requests coming from EC2 instances: see https://serverfault.com/questions/638096/how-is-cloudformation-metadata-access-controlled/787234 and https://stackoverflow.com/questions/29393675/does-aws-cfn-init-need-a-profile-role-for-describestackresource/38112522). CloudFormation also allows the regular instance-based credentials calls (with the `--role` option passed to `cfn-init`) but that is not how Beanstalk uses `cfn-init` so adding the correct permission to the instance role does not help.

The issue is that, when it gets launched by AutoSpotting, a spot instance does not have the `aws:cloudformation:stack-id` tag and is not a part of the CloudFormation stack. The tag cannot be added to an instance by AutoSpotting itself as tags starting with `aws:` are restricted to AWS services. The tag gets added once the instance gets attached to the auto-scaling group but it is too late for `cfn-init` to run properly.
[...]

Issue #343, opened 2019-05-15
Detect configuration drift between the group and the new spot instance when attaching it to the group


cristim:
# Github issue #

## Issue type ##

- Feature Idea

## Summary ##

It was reported that Beanstalk environments sometimes detect version mismatch between the new spot instance and other existing instances. 

Before attaching the spot instance, we should check if the instance configuration matches the other instances in the group.

If the instance is no longer identical to the group
- try to patch the configuration 
- if we fail patching it, we should terminate the instance instead of adding it to the group.  

## Steps to reproduce ##

- Create a Beanstalk environment
- Run AutoSpotting to replace an instance 
- Deploy a new version on Beanstalk
- Run AutoSpotting again when the instance is out of the grace period, in order to attach it to the group

 
## Expected results ##

- The instance should be updated to match the new group configuration, or terminated if we fail to update it.

## Actual results ##

- The instance will still run the previous version

cristim:
#354 is likely to make this obsolete, but we need to re-check once that's merged

cristim:
The recently merged event-based execution mode should address this problem for most cases.

Closing this issue.

Issue #332, opened 2019-03-22
AutoSpotting can cause client stack deletions to fail


iainelder:
## Issue type ##

- Bug Report

## Build number #

```
961
```

## Configuration ##

Default settings

## Environment ##

- Region: us-west-2
- Type of environment: DefaultVPC
- Anonymized launch configuration: Not relevant

## Summary ##

AutoSpotting can cause the deletion of a client stack to fail.

A client stack is any whose ASG has the spot-enabled=true tag.

When a client stack is deleted, one of the security groups might fail to be deleted because it is a dependency of a recently-launched spot instance not yet attached to the group.

To fix it, the dependent spot instance needs to be deleted, and then the stack deletion needs to be repeated.

## Steps to reproduce ##

The "recently-launched" part is the key. To reproduce it, I think this has to happen:

* AutoSpotting launches a new spot instance
* You delete the stack before the instance's grace period has elapsed
* CloudFormation deletes the ASG, terminating all instances in the ASG, leaving intact the new spot instance
* CloudFormation deletes the security group for the instances in the ASG

## Expected results ##

That the client stack deletion terminates normally.

## Actual results ##

At the last step, the stack deletion should fail because the security group is still being used by new spot instance.

The CloudFormation event log will show an error like this:

```
19 Mar 2019 19:37:39	WebServerSecurityGroup	DELETE_FAILED	resource sg-037aadff951412345 has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation; Request ID: 8fdc715e-9914-450d-81e7-9f80bc012345)
```

iainelder:
This is a tricky edge case.

It is most likely to affect development and testing environments, where stacks are set up and torn down regularly.

As it theoretically affects all client stacks, it would be great to have a standard solution.

I don't see any obvious way to solve it on the AutoSpotting side.

Maybe it can be solved with some sort of pre-delete hook on the client side.

cristim:
Maybe there is an event emitted when deleting Cloudformation stacks, which we can use to find such stale spot instances so we can terminate them.

We would find all the Autoscaling groups belonging to the terminating stack (they are tagged with the stack name) then we can get the unattached spot instances launched for this group and we can terminate all of them, maybe after removing all the security groups from them.

This would likely work but it assumes there's such an event emitted when deleting a Cloudformation stack.

iainelder:
> Maybe there is an event emitted when deleting Cloudformation stacks

[...]

Issue #296, opened 2018-09-14
Investigate a workaround for the lack of launch Lifecycle Hooks


cristim:
# Github issue #

## Issue type ##

- Feature Idea

## Summary ##

The group lifecycle hooks are not triggered for the spot instances, because they are launched outside the group.

We should try to implement a workaround, perhaps by changing the lifecycle state manually using the `ec2.SetCode` API calls if the group has some lifecycle hooks defined.

We should be careful not to block the launch of instances on groups that don't have any lifecycle hooks configured.

## Steps to reproduce ##

- have a group with lifecycle hooks
- trigger an instance replacement

## Expected results ##

- the lifecycle hook would be triggered when launching the spot instance outside the group

## Actual results ##

- the lifecycle hook is not triggered

cristim:
Thanks @earzur for your comment, I totally agree that lifecycle hooks are often useful for many production use cases.

Currently we do support them out of the box when terminating the on demand instances, but unfortunately for the launched spot instances this is a bit tricky,and we don't have a solution for this at the moment.

If we attach the spot instances right away, like you suggested, they would increase the group capacity so that the Autoscaling group's scaling policies can take action and terminate the new spot instances, unless we protect them from termination for a while(maybe as much as the grace period) or temporarily suspend the scaling processes for the group. Did you test this behavior, I kind of doubt that the lifecycle hooks would run for these spot instances if we attach them immediately.

But I haven't yet tested any of these scenarios and so far I don't know of any way to achieve this without interfering too much with the Autoscaling group.

But by all means please try to test these things and if they hold true I'm very interested to see a PR that changes the current behavior and adds support for the lifecycle hooks for the spot instances as well.

cristim:
@earzur no problem, we can also make the PR, as long as you or someone else can hack the code enough to test and determine the actual behavior.

earzur:
i've been experimenting with this today and my main problem with our current setup (rancher hosts in an ASG, with a container that monitors scale-in events to deactivate the rancher host in the cluster) is that when autospotting takes the decision to move an on-demand instance to spot, it seems to remove the on-demand instance from the ASG before shutting it down. As the lifecycle hooks are specified in the ASG, they are never triggered for those, which doesn't allow to rebalance the containers in the host before it gets shutdown.

[...]

PR #354, opened 2019-07-16
Implement event-based instance replacement

cristim:
**Context**

This change attempts to replace on-demand instances with spot as soon as they
are launched (triggered by the on-demand instance entering the running state),
in order to avoid running expensive on-demand capacity in parallel with the
newly launched spot instances. This should cover new groups out of the box, as
well as scaling out or replacement of terminated spot instances.

It has a few nice side effects concerning cost, since on-demand instances are
likely to run for just a few seconds, as well as potentially tackling some
known issues with ElasticBeanstalk(#343, #344) and deletion of CloudFormation
stacks(#332).

In addition, it implements support for handling the start lifecycle hooks(#296) by 
similarly  intercepting the lifecycle hook completion API failure event received
from Cloudtrail, and will immediately replace any scaled out instances, partially
addressing #284.

The current/legacy replacement logic will still kick in when enabling
AutoSpotting for the first time on an existing group, as well as when spot
instances fail to be launched so the group must failover to on-demand capacity
in order to avoid downtime.

Once the legacy replacement logic is done on the current group, the group is
tagegd with a configuration tag that enables the new event-based logic. This tag
is also set for all groups newer than an hour so their instances are automatically
replaced by the event-based replacement logic, and can be used to manually revert
a group back to the legacy logic in case of problems with the event-based implementation.

**Current status**
- changed regional CloudFormation stack to intercept EC2 instances
  as soon as entering the 'pending' and 'running' state'
- implement handling code for these events, that launches a replacement spot
  instance for each on-demand instance currently in pending state and performs
  the usual instance swapping logic when a spot instance enters the running
  state, without waiting for it to exit its grace period.
- largely kept the initial instance replacement logic as it is (waiting for
  the grace period) to avoid introducing regressions, and also in case the
  events are not properly propagated like in case of misconfigurations.
- implemented a configuration override that allows users to disable the event-based logic on a per-group level by tagging the group. 
- this code was executed against against real infrastructure and is
  working as expected for a couple of groups I tested it against.
- still missing test coverage for the new code and some existing tests were
  crashing and had to be disabled for now
- implemented support for handling the AutoScaling Lifecycle hooks when
  replacing new on-demand instances in event-based execution mode based on 
 CloudTrail events. As a caveat the lifecycle hook handling will be simulated with
 a delay of up to 15 minutes, because of the CloudTrail event propagation latency. 
- added support for running it locally against custom JSON event data in
  CloudFront Event format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html#ec2_event_type

**ToDo**
- test against more infrastructure
- iron out any potential crashes
- write automated tests for the new code and re-enable older test cases so
  that coverage doesn't drop
- port regional CloudFormation stack changes to the main template so
  it can also run from a StackSet
- update the documentation to mention this new behavior

# Issue Type

- Feature Pull Request
- Bugfix Pull Request
- Documentation Pull Request

   to it.
   and
   configurations.


core/autoscaling.go
```diff
@@ -75,7 +81,7 @@ func (a *autoScalingGroup) needReplaceOnDemandInstances() bool {
 				case DetachTerminationMethod:
 					randomSpot.terminate()
 				default:
-					a.terminateInstanceInAutoScalingGroup(randomSpot.Instance.InstanceId)
+					a.terminateInstanceInAutoScalingGroup(randomSpot.Instance.InstanceId, wait, false)
```
golangcibot: Error return value of `a.terminateInstanceInAutoScalingGroup` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -141,48 +152,98 @@ func (a *autoScalingGroup) process() {
...
+			spotInstanceID,
+			a.name)
 		return
 	}
 
 	logger.Println(a.region.name, "Found spot instance:", spotInstanceID,
 		"Attaching it to", a.name)
 
-	a.replaceOnDemandInstanceWithSpot(spotInstanceID)
+	a.replaceOnDemandInstanceWithSpot(nil, spotInstanceID)
```
golangcibot: Error return value of `a.replaceOnDemandInstanceWithSpot` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -191,10 +252,15 @@ func (a *autoScalingGroup) scanInstances() instances {
 	for _, inst := range a.Instances {
 		i := a.region.instances.get(*inst.InstanceId)
 
-		debug.Println(i)
-
 		if i == nil {
-			continue
+			debug.Println("Missing instance data for ", *inst.InstanceId, "scanning it again")
+			a.region.scanInstance(inst.InstanceId)
```
golangcibot: Error return value of `a.region.scanInstance` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -213,58 +279,52 @@ func (a *autoScalingGroup) scanInstances() instances {
...
+	if *a.DesiredCapacity == *a.MinSize {
+		attachErr := a.attachSpotInstance(spotInstanceID, true)
 		if attachErr != nil {
 			logger.Println(a.name, "skipping detaching on-demand due to failure to",
 				"attach the new spot instance", *spotInst.InstanceId)
 			return nil
 		}
 	} else {
-		defer a.attachSpotInstance(spotInstanceID)
+		defer a.attachSpotInstance(spotInstanceID, true)
```
golangcibot: Error return value of `a.attachSpotInstance` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -412,18 +472,34 @@ func (a *autoScalingGroup) setAutoScalingMaxSize(maxSize int64) error {
...
-		InstanceIds: []*string{
-			&spotInstanceID,
-		},
 	}
 
-	resp, err := svc.AttachInstances(&params)
+	// temporarily increase AutoScaling group in case it's of static size
+	if *a.MinSize == *a.MaxSize || *a.DesiredCapacity == *a.MaxSize {
+		logger.Println(a.name, "Temporarily increasing MaxSize")
+		a.setAutoScalingMaxSize(*a.MaxSize + 1)
```
golangcibot: Error return value of `a.setAutoScalingMaxSize` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -412,18 +472,34 @@ func (a *autoScalingGroup) setAutoScalingMaxSize(maxSize int64) error {
...
+		defer a.setAutoScalingMaxSize(*a.MaxSize)
```
golangcibot: Error return value of `a.setAutoScalingMaxSize` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -461,29 +550,68 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+			AutoScalingGroupName: a.AutoScalingGroupName,
+		})
+
+	if err != nil {
 		logger.Println(err.Error())
 		return err
 	}
 
+	for _, hook := range resDLH.LifecycleHooks {
+		asSvc.CompleteLifecycleAction(
```
golangcibot: Error return value of `asSvc.CompleteLifecycleAction` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -186,30 +187,85 @@ func (i *instance) canTerminate() bool {
...
+		return false
+	}
+
+	for _, asg := range i.region.enabledASGs {
+		if asg.name == *asgName && asg.isEnabledForEventBasedInstanceReplacement() {
+			asg.config = i.region.conf.AutoScalingConfig
+			asg.scanInstances()
+			asg.loadDefaultConfig()
+			asg.loadConfigFromTags()
+			asg.loadLaunchConfiguration()
```
golangcibot: Error return value of `asg.loadLaunchConfiguration` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -665,6 +750,108 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+	asgName := i.getReplacementTargetASGName()
+
+	if asgName == nil {
+		logger.Printf("Spot instance %s is missing ASG name", *i.InstanceId)
+		return nil, fmt.Errorf("spot instance %s is missing asg tag", *i.InstanceId)
+	}
+
+	if !odInstance.shouldBeReplacedWithSpot() {
+		logger.Printf("Target on-demand instance %s shouldn't be replaced", *odInstanceID)
+		i.terminate()
```
golangcibot: Error return value of `i.terminate` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -665,6 +750,108 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+
+	max := *asg.MaxSize
+
+	if *asg.DesiredCapacity == max {
+		if err := asg.setAutoScalingMaxSize(max + 1); err != nil {
+			logger.Printf("%s Couldn't temporarily expand ASG %s",
+				i.region.name, *asg.AutoScalingGroupName)
+			return nil, fmt.Errorf("couldn't increase ASG %s", *asg.AutoScalingGroupName)
+		}
+		defer asg.setAutoScalingMaxSize(max)
```
golangcibot: Error return value of `asg.setAutoScalingMaxSize` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -665,6 +750,108 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+	}
+
+	if err := asg.attachSpotInstance(*i.InstanceId, true); err != nil {
+		logger.Printf("Spot instance %s couldn't be attached to the group %s, terminating it...",
+			*i.InstanceId, asg.name)
+		i.terminate()
```
golangcibot: Error return value of `i.terminate` is not checked (from `errcheck`)

core/main.go
```diff
@@ -143,3 +167,133 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+
+	// If the event is for an Instance Spot Interruption
+	if cloudwatchEvent.DetailType == "EC2 Spot Instance Interruption Warning" {
+		log.Println("Triggered by", cloudwatchEvent.DetailType)
+		if instanceID, err := getInstanceIDDueForTermination(cloudwatchEvent); err != nil {
+			log.Println("Could't get instance ID of terminating spot instance", err.Error())
+			return
+		} else if instanceID != nil {
+			spotTermination := newSpotTermination(cloudwatchEvent.Region)
+			spotTermination.executeAction(instanceID, a.config.TerminationNotificationAction)
```
golangcibot: Error return value of `spotTermination.executeAction` is not checked (from `errcheck`)

core/main.go
```diff
@@ -143,3 +167,133 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+
+		// If event is Instance state change
+	} else if cloudwatchEvent.DetailType == "EC2 Instance State-change Notification" {
+		log.Println("Triggered by", cloudwatchEvent.DetailType)
+		instanceID, state, err := parseEventData(cloudwatchEvent)
+		if err != nil {
+			log.Println("Could't get instance ID of newly launched instance", err.Error())
+			return
+		} else if instanceID != nil {
+			a.handleNewInstanceLaunch(cloudwatchEvent.Region, *instanceID, *state)
```
golangcibot: Error return value of `a.handleNewInstanceLaunch` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -665,6 +750,108 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+
+	if !odInstance.shouldBeReplacedWithSpot() {
+		logger.Printf("Target on-demand instance %s shouldn't be replaced", *odInstanceID)
+		i.terminate()
+		return nil, fmt.Errorf("target instance %s should not be replaced with spot",
+			*odInstanceID)
+	}
+
+	asg := i.region.findEnabledASGByName(*asgName)
+	if asgName == nil {
```
golangcibot: nilness: impossible condition: non-nil == nil (from `govet`)

autospotting.go
```diff
@@ -4,218 +4,162 @@ import (
...
 func main() {
+
 	if os.Getenv("AWS_LAMBDA_FUNCTION_NAME") != "" {
 		lambda.Start(Handler)
+	} else if eventFile != "" {
+		parseEvent, err := ioutil.ReadFile(eventFile)
+		if err != nil {
+			log.Fatal(err)
+		}
+		Handler(nil, parseEvent)
```
golangcibot: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (from `staticcheck`)

core/autoscaling.go
```diff
@@ -141,48 +152,98 @@ func (a *autoScalingGroup) process() {
...
+			},
+		},
+	})
+	if err != nil {
+		logger.Println("Failed to enable ASG for event-based instance replacement:", err.Error())
+	}
+}
+
+func (a *autoScalingGroup) isEnabledForEventBasedInstanceReplacement() bool {
+	if time.Now().Sub(*a.CreatedTime) < time.Hour {
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

core/autoscaling.go
```diff
@@ -516,3 +644,12 @@ func (a *autoScalingGroup) alreadyRunningInstanceCount(
 	logger.Println(a.name, "Found", count, instanceCategory, "instances running on a total of", total)
 	return count, total
 }
+
+func (a *autoScalingGroup) isEnabled() bool {
```
golangcibot: U1000: func `(*autoScalingGroup).isEnabled` is unused (from `unused`)

core/instance_test.go
```diff
@@ -2243,9 +2189,147 @@ func Test_instance_isSameArch(t *testing.T) {
...
+		})
+	}
+}
+
+func Test_instance_isUnattachedSpotInstanceLaunchedForAnEnabledASG(t *testing.T) {
+
+	tests := []struct {
+		name           string
+		i              *instance
+		wantASG        *autoScalingGroup
```
golangcibot: U1000: field `wantASG` is unused (from `unused`)

core/main.go
```diff
@@ -65,11 +87,12 @@ func addDefaultFilter(cfg *Config) {
 	}
 }
 
-func disableLogging() {
-	setupLogging(&Config{LogFile: ioutil.Discard})
+func (cfg *Config) disableLogging() {
```
golangcibot: U1000: func `(*Config).disableLogging` is unused (from `unused`)

cristim:
For anyone brave enough to be a guinea pig, to install the latest build from this branch just click on this URL:
https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/create/template?stackName=AutoSpotting&templateURL=https://s3.amazonaws.com/cloudprowess/custom/template_build_1439509.yaml

core/instance.go
```diff
@@ -673,6 +739,120 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+	// Attach the new spot instance in parallel with the termination of the
+	// running on-demand instance in order to avoid increasing the capacity enough
+	// for the ASG to notice it and start terminating instances. The ASGs were
+	// sometimes seen to just terminates newly launched spot instances so the
+	// group was unintentionally kept with on-demand capacity.
+	go func() {
+		if err := asg.attachSpotInstance(*i.InstanceId, true); err != nil {
+			logger.Printf("Spot instance %s couldn't be attached to the group %s, terminating it...",
+				*i.InstanceId, asg.name)
+			i.terminate()
```
golangcibot: Error return value of `i.terminate` is not checked (from `errcheck`)

core/main.go
```diff
@@ -143,3 +167,256 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+		if err != nil {
+			log.Println("Could't get instance ID of newly launched instance", err.Error())
+			return
+		} else if instanceID != nil {
+			a.handleNewInstanceLaunch(cloudwatchEvent.Region, *instanceID, *state)
+		}
+
+	} else if eventType == "AWS API Call via CloudTrail" {
+		log.Println("Triggered by", eventType)
+		a.handleLifecycleHookEvent(cloudwatchEvent)
```
golangcibot: Error return value of `a.handleLifecycleHookEvent` is not checked (from `errcheck`)

core/action.go
```diff
@@ -0,0 +1,82 @@
...
+// terminates a random spot instance after enabling the event-based logic
+type terminateSpotInstance struct {
+	target target
+}
+
+func (tsi terminateSpotInstance) run() {
+	asg := tsi.target.asg
+
+	asg.enableForInstanceLaunchEventHandling()
+	asg.terminateRandomSpotInstanceIfHavingEnough(
```
golangcibot: Error return value of `asg.terminateRandomSpotInstanceIfHavingEnough` is not checked (from `errcheck`)

core/action.go
```diff
@@ -0,0 +1,82 @@
...
+	target target
+}
+
+func (tusi terminateUnneededSpotInstance) run() {
+	asg := tusi.target.asg
+	total := tusi.target.totalInstances
+	spotInstance := tusi.target.spotInstance
+	spotInstanceID := *spotInstance.InstanceId
+
+	asg.terminateRandomSpotInstanceIfHavingEnough(total, true)
```
golangcibot: Error return value of `asg.terminateRandomSpotInstanceIfHavingEnough` is not checked (from `errcheck`)

core/action.go
```diff
@@ -0,0 +1,82 @@
...
+	logger.Println("Spot instance", spotInstanceID, "is not need anymore by ASG",
+		asg.name, "terminating the spot instance.")
+	spotInstance.terminate()
```
golangcibot: Error return value of `spotInstance.terminate` is not checked (from `errcheck`)

core/action.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+type swapSpotInstance struct {
+	target target
+}
+
+func (ssi swapSpotInstance) run() {
+	asg := ssi.target.asg
+	spotInstanceID := *ssi.target.spotInstance.InstanceId
+	asg.replaceOnDemandInstanceWithSpot(nil, spotInstanceID)
```
golangcibot: Error return value of `asg.replaceOnDemandInstanceWithSpot` is not checked (from `errcheck`)

core/action.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (lsr launchSpotReplacement) run() {
+	spotInstanceID, err := lsr.target.onDemandInstance.launchSpotReplacement()
+	if err != nil {
+		logger.Printf("Could not launch cheapest spot instance: %s", err)
+		return
+	}
+	logger.Printf("Successfully launched spot instance %s, exiting...", *spotInstanceID)
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

core/autoscaling_test.go
```diff
@@ -2505,750 +2884,1404 @@ func TestGetDisallowedInstanceTypes(t *testing.T) {
...
+			if got := tt.asg.cronEventAction(); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("autoScalingGroup.cronEventAction() \ngot %#v, \nwant %#v", got, tt.want)
+			}
+			as.hourlySavings = 0
+		})
+	}
+}
 
+func Test_autoScalingGroup_enableForInstanceLaunchEventHandling(t *testing.T) {
+	type fields struct {
```
golangcibot: U1000: type `fields` is unused (from `unused`)

mello7tre:
Hi, i begin porting regional CloudFormation stack changes to the main template so new event-based replacement can be deployed using StackSet too.

As soon i found some time to test it i will notify you.
(I will try to test not only stackset creation but the whole new event-based replacement.)

mello7tre:
First test on deploying using Stackset was successfully.

What follow is the problem currently encountered testing event based replacement, but first let me tell that you have made a great work, i like the new code, in particular how all "events" are handled in a similar way (and are clearly detectable even in logs), this will allow to implement new features easily.

I began with a simple setup:
ASG without any scaling policies Min:0 Max:4 Desired: 1 no AutoSpot tag.
Once 1 instance was running i tagged it to use autospot and event based repl.
It replaced the running instance with the classic schedule replacement.

Then i increased the Desired to 2.
* OnDemand is launched
* OnDemand start event is correctly intercepted by AutoSpot (AS)
* Spot instance is launched
* Spot Start event is intercepted by AS 
* AS begin to replace previous launched OnDemand
  * Spot instance is attached to ASG
  * AS is unable to terminate OnDemand cause a "Scaling Activity is In Progress"

 i begin investing and found that the scaling activity is one "generated" by the new OnDemand launched.

Here the relevant AS logs:
```
2019/11/20 15:20:29 instance.go:789: Found unattached spot instance i-0b8df18b83d723d11
2019/11/20 15:20:29 main.go:396: eu-west-1 Found instance i-0b8df18b83d723d11 is not yet attached to its ASG, attempting to swap it against a running on-demand instance
2019/11/20 15:20:29 instance.go:248: eu-west-1 instace i-060e53d6462268e1d belongs to enabled ASG asg-autospot-test-AutoScalingGroup-1U1YTZ4XXEBCX
2019/11/20 15:20:29 instance.go:835: Attaching spot instance i-0b8df18b83d723d11 to the group asg-autospot-test-AutoScalingGroup-1U1YTZ4XXEBCX
2019/11/20 15:20:30 autoscaling.go:597: eu-west-1 asg-autospot-test-AutoScalingGroup-1U1YTZ4XXEBCX Terminating instance: i-060e53d6462268e1d
2019/11/20 15:20:30 autoscaling.go:631: ScalingActivityInProgress: Activity e615ba25-9913-4de8-f010-f4fbb40f591b is in progress.
2019/11/20 15:20:30 instance.go:847: On-demand instance i-060e53d6462268e1d couldn't be terminated, re-trying...
```

And the scaling activity involved:
```
{
    "Activities": [
        {
            "Description": "Launching a new EC2 instance: i-060e53d6462268e1d", 
            "AutoScalingGroupName": "asg-autospot-test-AutoScalingGroup-1U1YTZ4XXEBCX", 
            "ActivityId": "e615ba25-9913-4de8-f010-f4fbb40f591b", 
            "Details": "{\"Availability Zone\":\"eu-west-1b\"}", 
            "StartTime": "2019-11-20T15:20:11.219Z", 
            "Progress": 100, 
            "EndTime": "2019-11-20T15:20:42Z", 
            "Cause": "At 2019-11-20T15:20:02Z a user request update of AutoScalingGroup constraints to min: 0, max: 4, desired: 1 changing the desired capacity from 0 to 1.  At 2019-11-20T15:20:09Z an instance was started in response to a difference between desired and actual capacity, increasing the capacity from 0 to 1.", 
            "StatusCode": "Successful"
        }
    ]
}
```

As you can see the scaling activity end at 15:20:42 and AS try to terminate the instance at 15:20:30.

So we have now 2 running instances in the ASG with capacity desired "wrongly" set to 2.
As the two instances are identically and run in the same AZ, AWS try to balance the AZones launching a new instance.
```
availability zones eu-west-1c eu-west-1a had 0 2 instances respectively. An instance was launched to aid in balancing the group's zones.
```

this create a "loop" increasing the ASG to 5 running instances!

So i tried a simpy fix adding a `for cycle` to effectively re-try to terminate the instance.
`core/instance.go` method `swapWithGroupMember`
```
        retry := 1
        maxRetry := 5
        for retry < maxRetry {
                if err := asg.terminateInstanceInAutoScalingGroup(odInstanceID, true, true); err != nil {
                        logger.Printf("On-demand instance %s couldn't be terminated, re-trying...",
                                *odInstanceID)
                        time.Sleep(time.Duration(5 * retry) * time.Second)
                        retry += 1
                        continue
                } else {
                        return odInstance, nil
                }
        }

        return nil, fmt.Errorf("couldn't terminate on-demand instance %s",
                *odInstanceID)
```

this fixed the problem, and AS is now able to terminate the OnDemand instance.
But i do not know if this "delay" will be always sufficient.


Apart all this, i have one question:
why do you suspend termination ?
Before attaching Spot Instance to ASG you suspend termination for 5 minutes creating a defer waiter in swapWithGroupMember.
5 min is a long time and, as ASG launching new instance event can occur anytime, there could be a situation where a previous event can reactivate termination just after a following one disabled it (hope to be clear).
Let me do an example.
At 01:03 a new instance is launched, all goes well and AS replace it with Spot suspending termination until 01:08.
At 01:07:50 a new instance is launched, AS intercept the event and begin replacing it with a Spot, suspend termination but after few seconds AS Lambda execution of 01:03 re-enable it.

There is a reson why we cannot resume termination after having successfully terminated the OnDemand instance ?

I will do more testing, in particular i am concerned of ASG configured with a `CreationPolicy` and using a Policy `AutoScalingReplacingUpdate`  or `AutoScalingRollingUpdate`.
If a launched OnDemand is terminated before signaling success using cfn-signal can be problems.
I will try and report you.

Regards, Alberto.

cristim:
@mello7tre thank you, I really appreciate your contribution on this topic.

Honestly I didn't notice those failures that you implemented retries for, just their symptom of running more capacity than needed.

The suspended terminations was a workaround attempt, in order to avoid the group from interfering with the replacement logic, considering that we're trying to terminate the instances ourselves, but as you noticed, it didn't really help.

I guess if your retry logic makes it all work we can try to avoid suspending the termination, process, I'd love to avoid that if possible.

mello7tre:
I tried multiple time to be sure that was not caused by race condition or by some "strange" concurrency.
But the result is alway the same:
without the "retry" code it always happen.
Just for info:
i used a really simple "LaunchConfiguration" with only the imageId and a simple "User-Data" too.
```
    "LaunchConfiguration" : {
      "Type": "AWS::AutoScaling::LaunchConfiguration",
      "Properties": {
	"ImageId" : "ami-040ba9174949f6de4",
	"InstanceType"   : "t3.micro",
	"UserData"       : { "Fn::Base64" : { "Fn::Join" : ["", [
	  "#!/bin/bash\n",
	  "#", { "Ref" : "EnvStackVersion" }, "\n",
          "PATH=/opt/aws/bin:$PATH\n",
	  "yum install -y ec2-instance-connect\n",
          "sleep 420\n",
	  { "Fn::Sub": "cfn-signal -e $? --stack ${AWS::StackName} --resource AutoScalingGroup --region ${AWS::Region}\n" }
	]]}}
      }
    }
```

the region as you can see from the logs is eu-west-1.

But as i told you, this problem can be solved introducing the "retry" code.
The real BIG problem is the concurrency during increasing and decreasing the ASG CapacityMax.
I made the following test:
ASG configured with tags for autospotting using event based replacement:
0 Min - 0 Desired - 4 Max 

I increased Desired to 4.
Here the ASG activity logs:
```

Successful | Terminating EC2 instance: i-09b4903a0248ed3bd | At  2019-11-25T10:07:27Z an instance was taken out of service in response  to a difference between desired and actual capacity, shrinking the  capacity from 4 to 2.  At 2019-11-25T10:07:27Z instance  i-0d1238a395958db75 was selected for termination.  At  2019-11-25T10:07:28Z instance i-09b4903a0248ed3bd was selected for  termination. | 2019 November 25, 11:07:28 AM +01:00 | 2019 November 25, 11:08:50 AM +01:00
Successful | Terminating EC2 instance: i-0d1238a395958db75 | At  2019-11-25T10:07:27Z an instance was taken out of service in response  to a difference between desired and actual capacity, shrinking the  capacity from 4 to 2.  At 2019-11-25T10:07:27Z instance  i-0d1238a395958db75 was selected for termination.  At  2019-11-25T10:07:28Z instance i-09b4903a0248ed3bd was selected for  termination. | 2019 November 25, 11:07:27 AM +01:00 | 2019 November 25, 11:07:50 AM +01:00
Successful | Terminating EC2 instance: i-05d540274c5bf9117 | At  2019-11-25T10:06:36Z instance i-05d540274c5bf9117 was taken out of  service in response to a user request, shrinking the capacity from 3 to  2. | 2019 November 25, 11:06:36 AM +01:00 | 2019 November 25, 11:07:18 AM +01:00
Successful | Attaching an existing EC2 instance: i-042866082f8e62fd8 | At 2019-11-25T10:06:20Z an instance was added in response to user request. Keeping the capacity at the new 1. | 2019 November 25, 11:06:20 AM +01:00 | 2019 November 25, 11:06:30 AM +01:00
Successful | Terminating EC2 instance: i-067c429dde68c906e | At  2019-11-25T10:06:12Z instance i-067c429dde68c906e was taken out of  service in response to a user request, shrinking the capacity from 3 to  2. | 2019 November 25, 11:06:12 AM +01:00 | 2019 November 25, 11:07:14 AM +01:00
Successful | Attaching an existing EC2 instance: i-0d1238a395958db75 | At  2019-11-25T10:05:54Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 5 to 4.  At 2019-11-25T10:06:11Z an instance was added in response  to user request. Keeping the capacity at the new 1. | 2019 November 25, 11:06:11 AM +01:00 | 2019 November 25, 11:06:30 AM +01:00
Successful | Terminating EC2 instance: i-06023914f208fe79c | At  2019-11-25T10:06:10Z instance i-06023914f208fe79c was taken out of  service in response to a user request, shrinking the capacity from 3 to  2. | 2019 November 25, 11:06:10 AM +01:00 | 2019 November 25, 11:06:32 AM +01:00
Successful | Terminating EC2 instance: i-0d316ba87778e6d2a | At  2019-11-25T10:06:08Z instance i-0d316ba87778e6d2a was taken out of  service in response to a user request, shrinking the capacity from 4 to  3. | 2019 November 25, 11:06:08 AM +01:00 | 2019 November 25, 11:06:30 AM +01:00
Successful | Attaching an existing EC2 instance: i-0d30a9d03daed5f9a | At  2019-11-25T10:05:52Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 5 to 4.  At 2019-11-25T10:05:54Z an instance was added in response  to user request. Keeping the capacity at the new 1. | 2019 November 25, 11:05:54 AM +01:00 | 2019 November 25, 11:06:00 AM +01:00
Successful | Attaching an existing EC2 instance: i-09b4903a0248ed3bd | At 2019-11-25T10:05:52Z an instance was added in response to user request. Keeping the capacity at the new 1. | 2019 November 25, 11:05:52 AM +01:00 | 2019 November 25, 11:06:00 AM +01:00
Successful | Launching a new EC2 instance: i-067c429dde68c906e | At  2019-11-25T10:05:22Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 0 to 4.  At 2019-11-25T10:05:29Z an instance was started in  response to a difference between desired and actual capacity, increasing  the capacity from 0 to 4. | 2019 November 25, 11:05:31 AM +01:00 | 2019 November 25, 11:06:03 AM +01:00
Successful | Launching a new EC2 instance: i-06023914f208fe79c | At  2019-11-25T10:05:22Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 0 to 4.  At 2019-11-25T10:05:29Z an instance was started in  response to a difference between desired and actual capacity, increasing  the capacity from 0 to 4. | 2019 November 25, 11:05:31 AM +01:00 | 2019 November 25, 11:06:03 AM +01:00
Successful | Launching a new EC2 instance: i-0d316ba87778e6d2a | At  2019-11-25T10:05:22Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 0 to 4.  At 2019-11-25T10:05:29Z an instance was started in  response to a difference between desired and actual capacity, increasing  the capacity from 0 to 4. | 2019 November 25, 11:05:31 AM +01:00 | 2019 November 25, 11:06:03 AM +01:00
Successful | Launching a new EC2 instance: i-05d540274c5bf9117 | At  2019-11-25T10:05:22Z a user request update of AutoScalingGroup  constraints to min: 0, max: 4, desired: 4 changing the desired capacity  from 0 to 4.  At 2019-11-25T10:05:29Z an instance was started in  response to a difference between desired and actual capacity, increasing  the capacity from 0 to 4. | 2019 November 25, 11:05:31 AM +01:00 | 2019 November 25, 11:06:33 AM +01:00
```

as you can see the end result is a Desired Capacity of 2, instead of 4, and the relative termination of some launched spot instances.

So i tried to do some changes to better understood the source of the problem.
I created a method `waitForInstanceStatus` that wait until an instance is in an ASG.state (Ex InService) with a maximum retry times.

I inserted it in `instance.go` `swapWithGroupMember` before calling  asg.attachSpotInstance so that it wait that OnDemand instance finished Launching and his status is InService.

And before "returning" in `autoscaling.go` `attachSpotInstance` so that it wait that Spot instance finished attaching and his status is InService, before beginning to terminate the relative OnDemand.

This way i try to separate the time in which an OnDemand instance startup, the relative Spot is attached and the OnDemand is terminated.

I re-tried the same test:
This time the resulting Capacity Desired was even lower: 1.
But it's not fixed, depend on the concurrency, when i retried his final Capacity Desired was 3.

So the "real" problem is not the concurrency between start, attach and terminate actions.

I tried another time but with a different starting ASG "setup":
 0 Min - 0 Desired - 10 Max

This time all worked as expected, Spot replaced OnDemand and final ASG Capacity Desired was  4.

The problem lies in the concurrency of detecting ASG Desired and Max capacity.
Here what happen (for semplicity let's consider only two instances):
* Instances start near at the same time
* lambdas are triggered at the relative same time and launch spot instances
* lambdas are triggered
* one lambda check ASG max and desired and increase max by 1
* other lambda check ASG max and desired and increase max by 1
* as ASG data are retrieved as soon as the lambdas start (when they analyze the group) (and not rechecked before attaching the spot instances), the two lambda increase the ASG to the same max size (5 in our example).
* one lambda try to attach a spot and succeed
* other lambda try to attach a spot and get:
```
ValidationError: AutoScalingGroup asg-autospot-test-AutoScalingGroup-1H5EGE041FSI2 has min-size=0, max-size=5, and desired-size=5. To attach 1 instance, please update the AutoScalingGroup sizes appropriately.
```

But this is not the only problem during increasing/decreasing ASG Max Capacity.
When defer re-set previous Max Capacity it also use ASG data that are retrieved as soon as it started not the current one.
So while spot are attaching, desired capacity is greater than max capacity and ASG begin terminating instances, reducing capacity.
Then AutoSpotting begin terminating OnDemand, and final result is a lower Capacity Desired than expected.

So i created a method `changeAutoScalingMaxSize` that first get current Max Capacity and then increase/lower it for the value passed as parameter.
Re-tested all previous ASG setup, and always worked as expected.

But i am not 100% sure that this solve all concurrency problems.
The only solution would be to find a way to run Spot instances attach in a serial way.
(i tried to limit global AS lambda execution to only one, but i encountered other problems)

Sorry for the long comment, i will continue to work on this....

Regards, Alberto.

mello7tre:
Just for info, suspending termination for a long period is not a solution for the above problem (in fact it was already suspended when instances are terminated cause MaxSize < CapacityDesired), this is an extract of aws documentation about what processes are suspended:

Terminate
* Your Auto Scaling group does not scale in for alarms or scheduled actions that occur while the process is suspended. In addition, the following processes are disrupted:
  * AZRebalance is still active but does not function properly. It can launch new instances without terminating the old ones. This could cause your Auto Scaling group to grow up to 10 percent larger than its maximum size, because this is allowed temporarily during rebalancing activities. Your Auto Scaling group could remain above its maximum size until you resume the Terminate process. When Terminate resumes, AZRebalance gradually rebalances the Auto Scaling group if the group is no longer balanced between Availability Zones or if different Availability Zones are specified.
  * ReplaceUnhealthy is inactive but not HealthCheck. When Terminate resumes, the ReplaceUnhealthy process immediately starts running. If any instances were marked as unhealthy while Terminate was suspended, they will be immediately replaced.

mello7tre:
I was thinking:
we can create an sqs queue; in  `handleNewInstanceLaunch` instead of executing `swapWithGroupMember` we create a message in the queue with the instaceid of the Spot instance.
Then in the section code executed in the scheduled event, we parse the sqs queue and execute `swapWithGroupMember` for every instance returned.

This way we have the advantage of replacing instances during the same run of the scheduled replacing (hope to be clear).

Concurrency problems explained above will automatically solved.
There will be no further problem regarding concurrency of event replacing and scheduled one.

If i found some time i can begin working on this....

cristim:
@mello7tre how is it going, do you have any more issues to iron out?

This has been taking forever and i'm inclined to merge this as it is so we can get it better tested by our Alpha/Beta testers. Then we can iterate on it to mature it further.

Do you have anything major left to report at this point?

mello7tre:
i have a working version using the sqs queue, first tests are successful and i am no more experiencing previous problems.
i would like to do some more tests  (and then use it in my dev environment), but unfortunately do not have much time.

The logic is more or less this:
lambda triggered by new instance start event, if instance is a spot, instead of replacing ondemand with launched spot, insert in a sqs queue the following infos:
region asgName spotInstanceId.
lambda triggered by scheduled event before forking to process regions (in core/main.go method `processRegions`) get messages from queue (then delete them) and populate a new region property with this list.
Later, before Finding spot instance created by ASG, (in core.autoscaling.go method `cronEventAction`), lambda look up that list and search if there are spot instances for that region and asgName, in that case use them to replace ondemand ones. 
Then continue with "normal" schedule.

This way messages from sqs queue are retrieved only once for run, minimizing costs.
Ondemand replacement with spot instances from queue is executed in "serial" just before "normal" scheduled looking for a spot instance.

There is some more "tuning" to do on sqs ReceiveMessage's  WaitTimeSeconds property, leaving it as default (0) short pulling is used and if there are few messages only one is returned.
I am doing some test with a value of 5 seconds (with only two message in the queue) and sometimes it get all two sometimes only one, so maybe should be better to increase it some more.

If you want i can push my changes, but i would like to have some more time to clean the code a little and make a git comment that explain what i have done.
At the moment, i was in rush to write code and test it that i do not have properly commented the commits.

Just for info here the summary of the changes:
``` autospotting.go                                  |   1 +
 cloudformation/stacks/AutoSpotting/template.yaml |  14 +++++++++++
 core/autoscaling.go                              | 123 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---
 core/config.go                                   |   3 +++
 core/connections.go                              |   7 +++++-
 core/instance.go                                 |  40 ++++++++++++++++++++---------
 core/main.go                                     | 141 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------
 core/region.go                                   |  23 +++++++++++++++--
 8 files changed, 310 insertions(+), 42 deletions(-)
```



Tell me what you prefer...

Regards, Alberto.

cristim:
If you want to test it more that's fine, I'm just thinking that we have many users who can uncover more problems.

Regarding the SQS queue, why not trigger the Lambda when receiving messages from the queue? This would have the on demand instances replaced immediately.

mello7tre:
> If you want to test it more that's fine, I'm just thinking that we have many users who can uncover more problems.
> 
> Regarding the SQS queue, why not trigger the Lambda when receiving messages from the queue? This would have the on demand instances replaced immediately.

Because if multiple lambdas are triggered  in a short time, replacing ondemand on the same ASG,  we would still have the same problems of before regarding concurrency..

Maybe we can use sqs FIFO with a batch size of 1
https://aws.amazon.com/blogs/compute/new-for-aws-lambda-sqs-fifo-as-an-event-source/

but i should do some changes and more testing...

mello7tre:
> > If you want to test it more that's fine, I'm just thinking that we have many users who can uncover more problems.
> > Regarding the SQS queue, why not trigger the Lambda when receiving messages from the queue? This would have the on demand instances replaced immediately.
> 
> Because if multiple lambdas are triggered in a short time, replacing ondemand on the same ASG, we would still have the same problems of before regarding concurrency..
> 
> Maybe we can use sqs FIFO with a batch size of 1
> https://aws.amazon.com/blogs/compute/new-for-aws-lambda-sqs-fifo-as-an-event-source/
> 
> but i should do some changes and more testing...

Ideally we should try to achieve this:
process all events in parallel as soon as they are triggered, apart the one belonging to the same ASG, those events must be processed in a serial way (in the best approach as soon as they come).

Currently i do not have an "elegant" solution to do this.

But the root problem, that force us to use a serial method, is the increase of ASG Max Capacity when `capacity desired + spot instance to attach > capacity max` and the relative successive decrease.
If we find a way to overcome it we can go ahead with events/lambda full concurrency without problems.

What remain can be problems of concurrency with the scheduled event, but we can try to handle them.

mello7tre:
> > > If you want to test it more that's fine, I'm just thinking that we have many users who can uncover more problems.
> > > Regarding the SQS queue, why not trigger the Lambda when receiving messages from the queue? This would have the on demand instances replaced immediately.
> > 
> > 
> > Because if multiple lambdas are triggered in a short time, replacing ondemand on the same ASG, we would still have the same problems of before regarding concurrency..
> > Maybe we can use sqs FIFO with a batch size of 1
> > https://aws.amazon.com/blogs/compute/new-for-aws-lambda-sqs-fifo-as-an-event-source/
> > but i should do some changes and more testing...
> 
> Ideally we should try to achieve this:
> process all events in parallel as soon as they are triggered, apart the one belonging to the same ASG, those events must be processed in a serial way (in the best approach as soon as they come).
> 
> Currently i do not have an "elegant" solution to do this.
> 
> But the root problem, that force us to use a serial method, is the increase of ASG Max Capacity when `capacity desired + spot instance to attach > capacity max` and the relative successive decrease.
> If we find a way to overcome it we can go ahead with events/lambda full concurrency without problems.
> 
> What remain can be problems of concurrency with the scheduled event, but we can try to handle them.

I was thinking:
we can create another lambda with concurrency set to 1, we use as payload parameters:
* region
* ASG name
* value do increase/decrease ASG max size

Only when the main lamba try to increase the ASG max size and have an `Error.Code = "ValidationError"` (or we can search for string "To attach 1 instance, please update the AutoScalingGroup sizes appropriately" to be more confident), we increase the ASG calling synchronously the previous lambda and, if successful, we increase a counter too.
If we get an error `ErrCodeTooManyRequestsException` we simply retry.
When we finish to attach the instance and terminate the ondemand too, we execute again the lambda, to decrease the ASG by the value of our counter (best way is to use defer in swapWithGroupMember method).

What do you think of this ?
(it's a kind of  `try-catch` approach, we do not check is we need to increase the ASG, we simply try the operation and only if we get an error for not having the right max size, we increase it, but only `one lambda at time` can increase it.) 

This has the advantage to have full concurrency with operations on different ASG and to introduce a slightly delay only when we have the chances of conflicting operations.

mello7tre:
Have a working version using "external" lambda to "manage" asg max size.
Lambda is in python3.7 and code is inline in template and in the same region of "main" lambda, so deploy is identically.

First tests worked quite well.
Will continue to do more testing and will report you.

cristim:
Please upload the PR so we can get others to try it out. I will then build a custom binary out of it to make it easier.

mello7tre:
> Please upload the PR so we can get others to try it out. I will then build a custom binary out of it to make it easier.

ok tomorrow i will do it, just give me time to push a commit with a comment that explain all changes implemented.

cristim:
Sure, take your time and thank you for your hard work on this.

mello7tre:
> Sure, take your time and thank you for your hard work on this.

Pushed draft PR
#371

cristim:
Closing this, I just merged #371

gabegorelick:
@cristim I've split off some of the smaller, isolated refactorings from this PR: #392 #393 #394 

Just trying to make this changeset more manageable and land whatever I can from it.

autospotting.go
```diff
@@ -4,218 +4,163 @@ import (
...
-// Configuration handling
-func (c *cfgData) initialize() {
-
-	c.parseCommandLineFlags()
-
-	data, err := ec2instancesinfo.Data()
-	if err != nil {
-		log.Fatal(err.Error())
-	}
-	c.InstanceData = data
```
gabegorelick: @cristim I don't see where this is populated anymore. Is it no longer necessary with these changes?

cristim: Looks like this logic was moved to core/main.go around line 40.

gabegorelick: Ah I see. My mistake. Ctrl-F turned up nothing but that's because GitHub collapsed that file 🤦‍♂

mello7tre:
Hi, i found a problem in terminateRandomSpotInstanceIfHavingEnough logic that cause a loop of termination and running instance.
Consider the following setup:
ASG Min = Max = Desired = 2
ASG configured for Autospotting MinOnDemand = 1

Activate Autospotting, first only scheduled run.
After some time you will arrive at the following state:
1 Running onDemand
1 Running Spot.

Next time autspotting wil be executed it will enter a loop of termination and running.

Lets consider first cronEventAction, i will post only relevant lines:
 
```
spotInstance := a.findUnattachedInstanceLaunchedForThisASG()
 if spotInstance == nil {
        onDemandInstance := a.getAnyUnprotectedOnDemandInstance()
        if onDemandInstance == nil {
            return enableEventHandling{target{asg: a}}
         }
         if need, total := a.needReplaceOnDemandInstances(); !need {
           return terminateSpotInstance{target{asg: a, totalInstances: total}}
         }
}
```
as spotInstance is nil and onDemandInstance != nill we will call needReplaceOnDemandInstances, this will return false so we will end in:
terminateSpotInstance

this in return will call terminateRandomSpotInstanceIfHavingEnough

let's consider it (relevant lines too):
(totalRunning is total from previous methods)
```
func (a *autoScalingGroup) terminateRandomSpotInstanceIfHavingEnough(totalRunning int64, wait bool) error {
        if totalRunning == 1 {
                return nil
        }
        if a.allInstancesRunning() && a.instances.count64() < *a.DesiredCapacity {
                logger.Println("Not enough capacity in the group")
                return nil
        }
-------> after this point a randomly-selected spot instance is terminated
        randomSpot := a.getAnySpotInstance()
        if randomSpot == nil {
                return nil
        }
        logger.Println("Terminating randomly-selected spot instance",
                *randomSpot.Instance.InstanceId)
}

```
In our example we have a.instances.count64() = 2 and a.DesiredCapacity = 2 too, so even if all instances are running (a.allInstancesRunning() return true) we do not enter in second condition and we go on to terminating a random spot instance.
So in our example the only spot instance is terminated, ASG launch a new one, after some autopsotting cycle we return to our starting point (1 ondemand and 1 spot) and we re-enter the loop.
(loop is even faster if we have event-based replacement enabled)

I fixed it changing allInstancesRunning to return number of running ondemand too and changing terminateRandomSpotInstanceIfHavingEnough in the folliwing way:
```
func (a *autoScalingGroup) terminateRandomSpotInstanceIfHavingEnough(totalRunning int64, wait bool) error {

        if totalRunning == 1 {
                logger.Println("Warning: blocking replacement of very last instance - consider raising ASG to >= 2")
                return nil
        }

        if allInstancesAreRunning, onDemandRunning := a.allInstancesRunning(); allInstancesAreRunning {
                if a.instances.count64() == *a.DesiredCapacity && onDemandRunning == a.minOnDemand {
                        logger.Println("Currently Spot running equals to the required number, skipping termination")
                        return nil
                }

                if a.instances.count64() < *a.DesiredCapacity {
                        logger.Println("Not enough capacity in the group")
                        return nil
                }
        }
```
I made a test and now seems to work as expected.
Please tell me if you see some logic flaws and if not if you want me to push the PR with the fix.

cristim:
@mello7tre good catch, that looks good.

core/autoscaling.go
```diff
@@ -213,58 +310,60 @@ func (a *autoScalingGroup) scanInstances() instances {
...
-		attachErr := a.attachSpotInstance(spotInstanceID)
-		if attachErr != nil {
-			logger.Println(a.name, "skipping detaching on-demand due to failure to",
-				"attach the new spot instance", *spotInst.InstanceId)
-			return nil
-		}
-	} else {
-		defer a.attachSpotInstance(spotInstanceID)
+
+	a.suspendResumeProcess(*spotInst.InstanceId+"S", "suspend")
```
golangcibot, line 349: Error return value of `a.suspendResumeProcess` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -213,58 +310,60 @@ func (a *autoScalingGroup) scanInstances() instances {
...
+	defer a.suspendResumeProcess(*spotInst.InstanceId+"S", "resume")
```
golangcibot, line 350: Error return value of `a.suspendResumeProcess` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -213,58 +310,60 @@ func (a *autoScalingGroup) scanInstances() instances {
...
+
+	increase, attachErr := a.attachSpotInstance(*spotInst.InstanceId, true)
+	if increase > 0 {
+		defer a.changeAutoScalingMaxSize(int64(-1*increase), *spotInst.InstanceId)
```
golangcibot, line 354: Error return value of `a.changeAutoScalingMaxSize` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -673,6 +757,101 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+		logger.Printf("Target on-demand instance %s shouldn't be replaced", *odInstanceID)
+		i.terminate()
+		return nil, fmt.Errorf("target instance %s should not be replaced with spot",
+			*odInstanceID)
+	}
+
+	// var waiter sync.WaitGroup
+	// defer waiter.Wait()
+	// go asg.temporarilySuspendTerminations(&waiter)
+	asg.suspendResumeProcess(*i.InstanceId, "suspend")
```
golangcibot, line 824: Error return value of `asg.suspendResumeProcess` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -673,6 +757,101 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+	defer asg.suspendResumeProcess(*i.InstanceId, "resume")
```
golangcibot, line 825: Error return value of `asg.suspendResumeProcess` is not checked (from `errcheck`)

core/instance.go
```diff
@@ -673,6 +757,101 @@ func (i *instance) generateTagsList() []*ec2.TagSpecification {
...
+
+	logger.Printf("Attaching spot instance %s to the group %s",
+		*i.InstanceId, asg.name)
+	increase, err := asg.attachSpotInstance(*i.InstanceId, true)
+	if increase > 0 {
+		defer asg.changeAutoScalingMaxSize(int64(-1*increase), *i.InstanceId)
```
golangcibot, line 831: Error return value of `asg.changeAutoScalingMaxSize` is not checked (from `errcheck`)

core/autoscaling.go
```diff
@@ -331,6 +435,40 @@ func (a *autoScalingGroup) hasMemberInstance(inst *instance) bool {
...
+
+			if err != nil {
+				logger.Println(err.Error())
+				continue
+			}
+
+			autoScalingInstances := result.AutoScalingInstances
+
+			if len(autoScalingInstances) > 0 && *autoScalingInstances[0].LifecycleState == status {
+				isInstanceInStatus = true
```
golangcibot, line 459: ineffectual assignment to `isInstanceInStatus` (from `ineffassign`)

core/autoscaling.go
```diff
@@ -331,6 +435,40 @@ func (a *autoScalingGroup) hasMemberInstance(inst *instance) bool {
 	return false
 }
 
+func (a *autoScalingGroup) waitForInstanceStatus(instanceID *string, status string, maxRetry int) error {
+	isInstanceInStatus := false
+	for retry := 1; isInstanceInStatus == false; retry++ {
```
golangcibot, line 440: S1002: should omit comparison to bool constant, can be simplified to `!isInstanceInStatus` (from `gosimple`)

core/autoscaling.go
```diff
@@ -412,32 +550,144 @@ func (a *autoScalingGroup) setAutoScalingMaxSize(maxSize int64) error {
...
+	})
+
+	changed := false
+	seed := a.getRandSeed(instanceId)
+	svc := a.region.services.lambda
+
+	logger.Printf("Changing AutoScalingGroup %s MaxSize of %v unit",
+		a.name, value)
+
+	for retry, maxRetry := 0, 5; changed == false; {
```
golangcibot, line 585: S1002: should omit comparison to bool constant, can be simplified to `!changed` (from `gosimple`)

core/autoscaling.go
```diff
@@ -412,32 +550,144 @@ func (a *autoScalingGroup) setAutoScalingMaxSize(maxSize int64) error {
...
+		if err != nil {
+			logger.Printf("Issue while waiting for instance %v to start: %v",
+				spotInstanceID, err.Error())
+		}
+
+	}
+
+	increase := 0
+
+	for attaching := false; attaching == false; {
```
golangcibot, line 635: S1002: should omit comparison to bool constant, can be simplified to `!attaching` (from `gosimple`)

core/autoscaling.go
```diff
@@ -461,58 +711,234 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+	})
+
+	changed := false
+	seed := a.getRandSeed(instanceId)
+	svc := a.region.services.lambda
+
+	logger.Printf("Process %s for AutoScalingGroup %s",
+		action, a.name)
+
+	for retry, maxRetry := 0, 5; changed == false; {
```
golangcibot, line 912: S1002: should omit comparison to bool constant, can be simplified to `!changed` (from `gosimple`)

core/main.go
```diff
@@ -143,3 +167,277 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+		i.region.name, *i.InstanceId, *i.State.Name)
+
+	if state != "running" {
+		logger.Printf("%s Instance %s is not in the running state",
+			i.region.name, *i.InstanceId)
+		return errors.New("instance not in running state")
+	}
+
+	// Try OnDemand
+	if err := a.handleNewOnDemandInstanceLaunch(r, i); err != nil {
```
golangcibot, line 369: copylocks: call of a.handleNewOnDemandInstanceLaunch copies lock value: github.com/AutoSpotting/AutoSpotting/core.region contains sync.WaitGroup contains sync.noCopy (from `govet`)

core/main.go
```diff
@@ -143,3 +167,277 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+		return err
+	}
+
+	// Try Spot
+	if err := a.handleNewSpotInstanceLaunch(r, i); err != nil {
```
golangcibot, line 374: copylocks: call of a.handleNewSpotInstanceLaunch copies lock value: github.com/AutoSpotting/AutoSpotting/core.region contains sync.WaitGroup contains sync.noCopy (from `govet`)

core/main.go
```diff
@@ -143,3 +167,277 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+		return err
+	}
+	return nil
+}
+
+func (a *AutoSpotting) handleNewOnDemandInstanceLaunch(r region, i *instance) error {
```
golangcibot, line 380: copylocks: handleNewOnDemandInstanceLaunch passes lock by value: github.com/AutoSpotting/AutoSpotting/core.region contains sync.WaitGroup contains sync.noCopy (from `govet`)

core/main.go
```diff
@@ -143,3 +167,277 @@ func getRegions(ec2conn ec2iface.EC2API) ([]string, error) {
...
+	} else {
+		logger.Printf("%s skipping instance %s: either doesn't belong to an "+
+			"enabled ASG or should not be replaced with spot, ",
+			i.region.name, *i.InstanceId)
+		debug.Printf("%#v", i)
+	}
+	return nil
+}
+
+func (a *AutoSpotting) handleNewSpotInstanceLaunch(r region, i *instance) error {
```
golangcibot, line 399: copylocks: handleNewSpotInstanceLaunch passes lock by value: github.com/AutoSpotting/AutoSpotting/core.region contains sync.WaitGroup contains sync.noCopy (from `govet`)

core/autoscaling.go
```diff
@@ -461,58 +711,234 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+func (a *autoScalingGroup) isEnabled() bool {
+	for _, asg := range a.region.enabledASGs {
+		if asg.name == a.name {
+			return true
+		}
+	}
+	return false
+}
+
+func (a *autoScalingGroup) temporarilySuspendTerminations(wg *sync.WaitGroup) {
```
golangcibot, line 875: func `(*autoScalingGroup).temporarilySuspendTerminations` is unused (from `unused`)

core/autoscaling.go
```diff
@@ -461,58 +711,234 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+	if a.isTerminationSuspended() {
+		return
+	}
+
+	a.suspendTerminationProcess()
+	time.Sleep(300 * time.Second * a.region.conf.SleepMultiplier)
+	a.resumeTerminationProcess()
+}
+
+func (a *autoScalingGroup) isTerminationSuspended() bool {
```
golangcibot, line 888: func `(*autoScalingGroup).isTerminationSuspended` is unused (from `unused`)

core/autoscaling.go
```diff
@@ -461,58 +711,234 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+		&autoscaling.ScalingProcessQuery{
+			AutoScalingGroupName: a.AutoScalingGroupName,
+			ScalingProcesses:     []*string{aws.String("Terminate")},
+		})
+	if err != nil {
+		logger.Printf("couldn't suspend termination processes on ASG %s ", a.name)
+	}
+}
+
+func (a *autoScalingGroup) resumeTerminationProcess() {
```
golangcibot, line 854: func `(*autoScalingGroup).resumeTerminationProcess` is unused (from `unused`)

core/autoscaling.go
```diff
@@ -461,58 +711,234 @@ func (a *autoScalingGroup) detachAndTerminateOnDemandInstance(
...
+					count++
+				}
+			}
 		}
 	}
 	logger.Println(a.name, "Found", count, instanceCategory, "instances running on a total of", total)
 	return count, total
 }
+
+func (a *autoScalingGroup) suspendTerminationProcess() {
```
golangcibot, line 834: func `(*autoScalingGroup).suspendTerminationProcess` is unused (from `unused`)

mello7tre:
i know that this PR and the comment related to it are already a lot, but i found a bug and didn't want to open an issue as this is not have been yet merged in the master branch.
So i will explain it here, and then when i will be sure that the solution that i implemented is fine i will wil open a specific PR.

The problem is difficult to explain, but is revelant so i will try:
## Setup ##
Consider an ASG with a Desired = 10 and a AS minOnDemand = 1
Over time we will arrive at the following state:
9 Spot and 1 OnDemand
Now if the ASG can scale based on load Autoscaling can terminate an instance and reduce the Desired to 9.
Let's take the case where the terminated one is the only one OnDemand.
We will have 9 Spot and 0 OnDemand.
On the next run AS (AutoSpotting)will terminate a spot, so that a new onDemand is launched, to obey the condition of minOnDemand = 1.
Buy if the ASG have a lifecyclehook triggered on instance terminating and that lifecyclehook take more time than AS scheduled frequency, a big issue can arise.

## Issue ##
AS terminate spot instance(minOnDemand), instance is not terminated (LifeCycleHook) immediately, but Autoscaling Launch a new instance (onDemand).
On the next scheduled run, AS (cronEventAction):
```
                if need, total := a.needReplaceOnDemandInstances(); !need {
                        logger.Printf("Not allowed to replace any more of the running OD instances in %s", a.name)
                        return terminateSpotInstance{target{asg: a, totalInstances: total}}
                }

``` 
look if new onDemand must be replaced, but as `onDemandRunning == a.minOnDemand` need = false and he execute:
`terminateSpotInstance` that finally execute `terminateRandomSpotInstanceIfHavingEnough`

```
func (a *autoScalingGroup) terminateRandomSpotInstanceIfHavingEnough(totalRunning int64, wait bool) error {

        if totalRunning == 1 {
                logger.Println("Warning: blocking replacement of very last instance - consider raising ASG to >= 2")
                return nil
        }

        if allInstancesAreRunning, onDemandRunning := a.allInstancesRunning(); allInstancesAreRunning {
                if a.instances.count64() == *a.DesiredCapacity && onDemandRunning == a.minOnDemand {
                        logger.Println("Currently Spot running equals to the required number, skipping termination")
                        return nil
                }

                if a.instances.count64() < *a.DesiredCapacity {
                        logger.Println("Not enough capacity in the group")
                        return nil
                }
        }

        randomSpot := a.getAnySpotInstance()
        if randomSpot == nil {
                logger.Println("Couldn't pick a random spot instance")
                return nil
        }

        logger.Println("Terminating randomly-selected spot instance",
                *randomSpot.Instance.InstanceId)

        switch a.config.TerminationMethod {
        case DetachTerminationMethod:
                return randomSpot.terminate()
        default:
                return a.terminateInstanceInAutoScalingGroup(randomSpot.Instance.InstanceId, wait, false)
        }
}

```

the root problem is that the terminating instance is still in state running!
So `if a.instances.count64() == *a.DesiredCapacity && onDemandRunning == a.minOnDemand {` is not true and AS terminate another spot instance (can even happen that try to terminate the same already terminating!)
Instance begin terminating (LifeCycleHook) and Autoscaling launch a new instance, but on the next run the same problem arise, so we can end in a loop of launching and then terminating instances.
(i have observed all this in realtime)

## Solution/Fix ##
When executing `terminateRandomSpotInstanceIfHavingEnough` we need to consider not the instance State but the Instance Autoscaling LifeCycle state.
I have added to autoScalingGroup struct two new properties:
```
        instancesOnDemandInService map[string]*instance
        instancesSpotInService     map[string]*instance
```

then created a func `populateASGInstancesInService` that create the previous map with the id of the instance inService as index and the instance herself as value.
I have then changed `terminateRandomSpotInstanceIfHavingEnough`:
```
func (a *autoScalingGroup) terminateRandomSpotInstanceIfHavingEnough(wait bool) error {
        onDemandRunning := int64(len(a.instancesOnDemandInService))
        totalRunning := int64(len(a.instancesOnDemandInService) + len(a.instancesSpotInService))

        if totalRunning == 1 {
                logger.Println("Warning: blocking replacement of very last instance - consider raising ASG to >= 2")
                return nil
        }

        if totalRunning < *a.DesiredCapacity {
                logger.Println("Not enough capacity in the group, skipping termination")
                return nil
        }

        if totalRunning == *a.DesiredCapacity && onDemandRunning == a.minOnDemand {
                logger.Println("Currently Spot running equals to the required number, skipping termination")
                return nil
        }
.....
......
......
```
And finally to avoid terminating a terminating in progress instance i changed `getAnySptInstance`:
```
func (a *autoScalingGroup) getAnySpotInstance() *instance {
        for _, i := range a.instancesSpotInService {
                return i
        }
        return nil
}
```
I have an AS version builded with this code running, and so far found no problem.
I will notify you if i found any.
If all goes well, if you like, i can open a PR with the changes/fixes.

Regards, Alberto.

mello7tre:
Thinking again on this:
as the problem can arise even when terminating onDemand instance, maybe a better solution can be to exclude instances if their ASG lifecycle state begin with terminating directly in `scanInstances` func ....

mello7tre:
made some test, the right solution is to exclude instances when their ASG lifecycle state begin with `Terminating`.
I will open a PR

PR #360, opened 2019-08-20
fix typo

codenoid:


cristim:
Good catch, thanks for your contribution and looking forward for more 😁

PR #371, opened 2019-12-10
Feat/event based instance replacement

mello7tre:
# Issue Type

- Feature Pull Request
- Bugfix Pull Request
## Summary

    * Improvements on event based instance replacement
    
    - Ported to StackSet deploy mode.
    
    - Fix to "ScalingActivityInProgress" if launched ondemand instance is
    terminated before going "inservice":
    
            Created method "waitForInstanceStatus", that wait, until a max
            retry (5), that an instance belonging to an ASG is in the desired
            status (InService).
            (Sleep time is 5*retry)
    
    - Fix to multiple problems caused by lambda concurrency changing ASG
    MaxSize:
            Created another Lambda (LambdaManageASG) in the same region of
            the main one; code python3.7, inline in template.
            Lambda concurrency is set to one.
            Is function is to change ASG MaxSize by the ammount specified.
    
            Used a "try-catch approach":
            if AttachInstances return error code "ValidationError" and
            string "update the AutoScalingGroup sizes" is present in error
            message, means that we need to increase ASG MaxSize.
            So we execute method changeAutoScalingMaxSize that invoke
            LambdaManageASG.
    
            If invoke return error "ErrCodeTooManyRequestsException", means
            that multiple Main lambdas are executing LambdaManageASG, we
            sleep for a random interval part of seconds and retry.
    
            Method attachSpotInstance now return an int that represent the
            ammount of change to the ASG MaxSize, so that in
            swapWithGroupMember we can defer the call to
            changeAutoScalingMaxSize to decrease the ASG MaxSize of the
            previously increased ammount.
    
            We use "waitForInstanceStatus" in attachSpotInstance before
            returning to be sure that spot instance has been attached and
            are InService before beginning to terminate ondemand one.

   to it.
   and
   configurations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine as golang
+FROM golang:1.13-alpine as golang
 RUN apk add -U --no-cache ca-certificates git make
 COPY . /src
 WORKDIR /src
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine
+FROM golang:1.13-alpine
 RUN apk add -U --no-cache ca-certificates git make zip
 
 COPY . /src
--- a/START.md
+++ b/START.md
@@ -189,10 +189,53 @@ be risky, please handle with care.
 
 ### For Elastic Beanstalk ###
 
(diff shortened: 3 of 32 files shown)


cristim:
I’ll merge this as it is to get it tested by the evaluation users.

PR #381, opened 2020-01-25
Support spot price buffer percentage of 0

### LINKED ISSUE #380: spot_price_buffer_percentage=0 is ignored
## Issue type ##

Bug Report

## Build number #

master

## Summary ##

If `spot_price_buffer_percentage` is 0, the default of 10 is used. 

https://github.com/AutoSpotting/AutoSpotting/blob/9b438dc47cbfd9587b042d5b2aaac2765c28a4e0/core/autoscaling_configuration.go#L138-L141

But a value of 0 is useful to bid at the current spot price. 

My specific use case involves testing spot instance interruption. Bidding at the current spot price maximizes the chances of the instance being interrupted when the price rises, while still ensuring that we're able to allocate the instance in the first place. I can work around this by using a sufficiently small value of `spot_price_buffer_percentage`, but that's messier.

## Steps to reproduce ##

Set `spot_price_buffer_percentage=0` and bidding policy to `aggressive`.

## Expected results ##

Bid at current spot price.

## Actual results ##

Bid at 10% over spot price.

gabegorelick:
Fixes #380

# Issue Type

Bugfix Pull Request

## Summary

Bid at current spot price when spot_price_buffer_percentage is 0, instead of bidding at 10% above spot price.

See #380 for discussion.

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling_configuration.go
+++ b/core/autoscaling_configuration.go
@@ -135,7 +135,7 @@ func (a *autoScalingGroup) loadSpotPriceBufferPercentage(tagValue *string) (floa
 	if err != nil {
 		logger.Printf("Error with ParseFloat: %s\n", err.Error())
 		return DefaultSpotPriceBufferPercentage, false
-	} else if spotPriceBufferPercentage <= 0 {
+	} else if spotPriceBufferPercentage < 0 {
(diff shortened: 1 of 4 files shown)


gabegorelick:
I'm still working on testing this. Something seems to still not like using 0.

PR #386, opened 2020-01-26
Allow specifying GOOS and GOARCH

### LINKED ISSUE #385: `make build` forces GOOS=linux
# Github issue #

## Issue type ##

Bug Report

## Build number #

master

## Summary ##

The docs mention that `make build` makes an artifact suitable for running locally, but this is only true if you're on Linux. `make build` sets `GOOS=linux`, which is correct if you want to run the executable on Lambda, but not if you want to run it locally on a non-Linux platform.

https://github.com/AutoSpotting/AutoSpotting/blob/9b438dc47cbfd9587b042d5b2aaac2765c28a4e0/Makefile#L46-L48

## Steps to reproduce ##

1. Run `make build` on a platform other than Linux.
2. Run `./AutoSpotting`

## Expected results ##

`./AutoSpotting` runs without errors

## Actual results ##

`exec format error: ./AutoSpotting` since the executable is not compatible with the current architecture.

gabegorelick:
Fixes #385

# Issue Type

Feature Pull Request

   to it.
   and
   configurations.

### DIFF
--- a/CUSTOM_BUILDS.md
+++ b/CUSTOM_BUILDS.md
@@ -64,9 +64,12 @@ your CloudFormation or Terraform stack to use those new binaries.
 
 1. Build the code:
 
-   `make build` or `go build` should both work, the only difference is the
-   inclusion of versioning information into the binary, which may not matter
-   much anyway.
+   Run `make build` to generate a binary at `./AutoSpotting`. The resulting
+   binary is suitable for running on AWS Lambda. If you want a binary for
+   running locally and your local environment is not linux/amd64, try running
+   the following:
+
+   `GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) make build`
 
 ## Running locally ##
 
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ BUCKET_NAME ?= cloudprowess
 FLAVOR ?= custom
 LOCAL_PATH := build/s3/$(FLAVOR)
 LICENSE_FILES := LICENSE THIRDPARTY
+GOOS ?= linux
+GOARCH ?= amd64
 
 SHA := $(shell git rev-parse HEAD | cut -c 1-7)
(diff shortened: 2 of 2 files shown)


CUSTOM_BUILDS.md
```diff
@@ -64,9 +64,12 @@ your CloudFormation or Terraform stack to use those new binaries.
 
 1. Build the code:
 
-   `make build` or `go build` should both work, the only difference is the
-   inclusion of versioning information into the binary, which may not matter
-   much anyway.
+   Run `make build` to generate a binary at `./AutoSpotting`. The resulting
```
gabegorelick: I removed the reference to `go build` since it's easier to only document one supported method to build binaries. I can add it back if you disagree.

CUSTOM_BUILDS.md
```diff
@@ -64,9 +64,12 @@ your CloudFormation or Terraform stack to use those new binaries.
...
+   binary is suitable for running on AWS Lambda. If you want a binary for
+   running locally and your local environment is not linux/amd64, try running
+   the following:
+
+   `GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) make build`
```
gabegorelick: We can add a new make target to make this easier, e.g. `build_local`. But punting on that for now.

Makefile
```diff
@@ -44,7 +46,7 @@ update_deps:                                                 ## Update all depen
 .PHONY: update_deps
 
 build: build_deps                                            ## Build the AutoSpotting binary
-	GOOS=linux go build -ldflags=$(LDFLAGS) -o $(BINARY)
+	GOOS=$(GOOS) GOARCH=$(GOARCH) go build -ldflags=$(LDFLAGS) -o $(BINARY)
```
gabegorelick: We didn't specify `GOARCH` before, but pretty sure we want to default to amd64.

PR #392, opened 2020-01-28
Make loadLaunchConfiguration return a launchConfiguration

gabegorelick:
It previously mutated the receiver, but did not return the loaded launchConfiguration.

This was pulled out of #354, with some additional error handling.

# Issue Type

Feature Pull Request

## Summary

Pulling out a small part of the big event-based-instance-replacement work. This part seems fairly inconsequential, but need to start somewhere :) 

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -25,16 +25,16 @@ type autoScalingGroup struct {
 	config              AutoScalingConfig
 }
 
-func (a *autoScalingGroup) loadLaunchConfiguration() error {
+func (a *autoScalingGroup) loadLaunchConfiguration() (*launchConfiguration, error) {
 	//already done
 	if a.launchConfiguration != nil {
-		return nil
+		return a.launchConfiguration, nil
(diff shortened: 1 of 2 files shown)


core/autoscaling.go
```diff
@@ -25,16 +25,17 @@ type autoScalingGroup struct {
...
 	//already done
 	if a.launchConfiguration != nil {
-		return nil
+		return a.launchConfiguration, nil
 	}
 
 	lcName := a.LaunchConfigurationName
 
 	if lcName == nil {
-		return errors.New("missing launch configuration")
```
gabegorelick: Correct me if I'm wrong, but this doesn't seem like an error. I think it would return an error for all ASGs using LaunchTemplates. I believe this only worked since we were ignoring any errors returned by `loadLaunchConfiguration`.

cristim: I agree, good catch!

PR #393, opened 2020-01-28
Refactor instance.isProtectedFromTermination

gabegorelick:
Split off from #354 with some small tweaks to get tests to pass.

# Issue Type

Feature Pull Request

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling_test.go
+++ b/core/autoscaling_test.go
@@ -1210,22 +1210,8 @@ func TestGetOnDemandInstanceInAZ(t *testing.T) {
 					},
 				},
 			),
-			az: aws.String("1b"),
-			expected: &instance{
-				Instance: &ec2.Instance{
-					InstanceId:        aws.String("ondemand-running"),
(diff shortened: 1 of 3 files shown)


core/autoscaling_test.go
```diff
@@ -1210,22 +1210,8 @@ func TestGetOnDemandInstanceInAZ(t *testing.T) {
 					},
 				},
 			),
-			az: aws.String("1b"),
-			expected: &instance{
```
cristim: Is this test case redundant? Why delete it?

gabegorelick: It wasn't deleted, the expected value changed to `nil`, because now if there's an error describing instance attributes, we assume the instance is protected. Previously, we defaulted to assuming it was not protected, and that behavior was enshrined in this test.

See https://github.com/AutoSpotting/AutoSpotting/pull/393/files#diff-fa23d3c4fb090da3bc9f71371d6e1d95R163 for the salient change. That's copied over verbatim from the event-based-instance-replacement branch.

cristim: Got it, thank you!

PR #399, opened 2020-01-28
Rename travisci make targets to ci

gabegorelick:
These targets are pretty much generic and do not need to run in TravisCI. 

I specifically have a use case where I am running some of them in a different CI provider, so invoking `make travisci` is awkward. And who knows, maybe in the future we'll all be using GitHub actions 😉  

# Issue Type

Feature Pull Request

## Summary
Rename the following make targets:
- `travisci-cover` -> `ci-cover`
- `travisci-checks` -> `ci-checks`
- `travisci` -> `ci`
- `travisci-docker` -> `ci-docker`

   to it.
   and
   configurations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,21 +107,21 @@ html-cover: test                                             ## Display coverage
 	@go tool cover -html=$(COVER_PROFILE)
 .PHONY: html-cover
 
-travisci-cover: html-cover                                   ## Test & generate coverage in the TravisCI format, fails unless executed from TravisCI
+ci-cover: html-cover                                         ## Test & generate coverage in the TravisCI format, fails unless executed from TravisCI
 ifdef COVERALLS_TOKEN
 	@goveralls -coverprofile=$(COVER_PROFILE) -service=travis-ci -repotoken=$(COVERALLS_TOKEN)
 endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -109,21 +109,21 @@ html-cover: test                                             ## Display coverage
 	@go tool cover -html=$(COVER_PROFILE)
 .PHONY: html-cover
 
-travisci-cover: html-cover                                   ## Test & generate coverage in the TravisCI format, fails unless executed from TravisCI
+ci-cover: html-cover                                         ## Test & generate coverage in the TravisCI format, fails unless executed from TravisCI
```
gabegorelick: Kept the reference to TravisCI in the comment, since goveralls does output in TravisCI format and it won't work unless run in TravisCI (where `COVERALLS_TOKEN` is defined).

cristim:
I  agree with this change, thanks!

There's also the build number variable which tends to be different for each of the CI providers. But I guess that's only relevant for the builds I release, if that's missing you just get the SHA prefix which might be enough

gabegorelick:
Fixed the merge conflicts. This is ready to go now.

PR #404, opened 2020-01-30
Move TestMain from spot_price_test.go to main_test.go

gabegorelick:
This change was pulled from #354.

I actually think TestMain needs to be rethought in light of issues like https://github.com/AutoSpotting/AutoSpotting/issues/358 (real code doesn't always initialize globals, so neither should tests), but punting on that until the larger changes from #354 are done. 

   to it.
   and
   configurations.

### DIFF
--- a/core/main_test.go
+++ b/core/main_test.go
@@ -5,12 +5,31 @@ package autospotting
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
 	"reflect"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 )
+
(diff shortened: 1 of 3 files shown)


core/main_test.go
```diff
@@ -5,13 +5,32 @@ package autospotting
...
+	"os"
 	"reflect"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
+func TestMain(m *testing.M) {
+	var logOutput io.Writer
```
gabegorelick: This is the implementation from https://github.com/AutoSpotting/AutoSpotting/pull/354/files#diff-592039bf1534bef34939e90696678d3eR16, minus all the stuff that doesn't make sense before the rest of the changes from #354 are merged.

Issue #376, opened 2020-01-25
Instance types "Unavailable in this Availability Zone" with erroneous price of 0


gabegorelick:
## Issue type ##

Bug Report

## Build number #

9b438dc

## Configuration ##

`allowed_instance_types: current`, but not sure that matters.

## Environment ##

- AWS region: us-east-1
- Type of environment: VPC

## Summary ##

AutoSpotting is reporting that certain spot instance types have a price of 0.

Here's what I see in the logs:
> instance.go:376: Comparing current type t2.small with price 0.023 with candidate t2.small with price 0

Since AutoSpotting [interprets a spot price of $0 as meaning the instance type is unavailable](https://github.com/AutoSpotting/AutoSpotting/blob/972bc61ac8c3315bdbb30245057b24ab74baa17a/core/instance.go#L207), you get a message saying "Unavailable in this Availability Zone" and AutoSpotting refuses to use that instance type.

The weird thing is that calling `describe-spot-price-history` indicates that the instance type does seem to exist.
```sh
aws ec2 describe-spot-price-history --instance-types t2.small --availability-zone us-east-1b --product-description Linux/UNIX
{
    "SpotPriceHistory": [
        {
            "AvailabilityZone": "us-east-1b",
            "InstanceType": "t2.small",
            "ProductDescription": "Linux/UNIX",
            "SpotPrice": "0.006900",
            "Timestamp": "2020-01-24T09:30:58.000Z"
        },
```

This is happening intermittently for a number of instances types. I have no issues manually creating spot requests with the same instance type and AZ (they are fulfilled without issue). 

## Steps to reproduce ##

Setting `allowed_instance_types: current` makes the issue more visible, since if you have an "Unavailable instance" AutoSpotting won't substitute a replacement instance type.

cristim:
Thanks for reporting this.

This price data is coming from the spot instance price history, and intermittent issues might indicate API throttling, especially considering that you seem to be  trying to run multiple instances of AutoSpotting in parallel from the same account.

gabegorelick:
> This price data is coming from the spot instance price history, and intermittent issues might indicate API throttling

It seems to be happening more consistently for some instance types than others. E.g. I saw it for a few hours straight with t2.medium while other instances returned data fine. So I'm not sure API throttling is the issue (or if it is, AWS is doing very strange throttling).

> you seem to be trying to run multiple instances of AutoSpotting in parallel from the same account.

[...]

PR #408, opened 2020-01-31
Use paginated version of DescribeSpotPriceHistory

gabegorelick:
This may help with #376.

# Issue Type

- Bugfix Pull Request

## Summary

Use `DescribeSpotPriceHistoryPages` instead of `DescribeSpotPriceHistory`.

   to it.
   and
   configurations.

### DIFF
--- a/core/mock_test.go
+++ b/core/mock_test.go
@@ -28,9 +28,11 @@ func CheckErrors(t *testing.T, err error, expected error) {
 type mockEC2 struct {
 	ec2iface.EC2API
 
-	// Describe Spot Price History
-	dspho   *ec2.DescribeSpotPriceHistoryOutput
-	dspherr error
+	// DescribeSpotPriceHistoryPages output
+	dsphpo []*ec2.DescribeSpotPriceHistoryOutput
+
+	// DescribeSpotPriceHistoryPages error
+	dsphperr error
 
 	// DescribeInstancesOutput
 	dio *ec2.DescribeInstancesOutput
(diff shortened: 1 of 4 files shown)


core/mock_test.go
```diff
@@ -59,13 +61,16 @@ type mockEC2 struct {
...
+	for i, page := range m.dsphpo {
+		f(page, i == len(m.dsphpo)-1)
+	}
+	return m.dsphperr
 }
 
 func (m mockEC2) DescribeInstancesPages(in *ec2.DescribeInstancesInput, f func(*ec2.DescribeInstancesOutput, bool) bool) error {
 	f(m.dio, true)
-	return nil
+	return m.diperr
```
gabegorelick: `diperr` is never used, so I think that's why no one noticed this.

core/mock_test.go
```diff
@@ -28,9 +28,11 @@ func CheckErrors(t *testing.T, err error, expected error) {
 type mockEC2 struct {
 	ec2iface.EC2API
 
-	// Describe Spot Price History
-	dspho   *ec2.DescribeSpotPriceHistoryOutput
-	dspherr error
+	// DescribeSpotPriceHistoryPages output
+	dsphpo []*ec2.DescribeSpotPriceHistoryOutput
```
gabegorelick: Rename `dspho` to `dsphpo` and make it an array of outputs. Each element in the array corresponds to a page.

core/mock_test.go
```diff
@@ -59,13 +61,16 @@ type mockEC2 struct {
...
-	return m.dspho, m.dspherr
+func (m mockEC2) DescribeSpotPriceHistoryPages(in *ec2.DescribeSpotPriceHistoryInput, f func(*ec2.DescribeSpotPriceHistoryOutput, bool) bool) error {
+	for i, page := range m.dsphpo {
+		f(page, i == len(m.dsphpo)-1)
+	}
+	return m.dsphperr
 }
 
 func (m mockEC2) DescribeInstancesPages(in *ec2.DescribeInstancesInput, f func(*ec2.DescribeInstancesOutput, bool) bool) error {
 	f(m.dio, true)
```
gabegorelick: It looks like all existing `Pages` tests only test with a single page? We should think about changing that.

core/mock_test.go
cristim: Ouch! Good catch!

PR #413, opened 2020-02-02
Move config loading out of main and add tests for it

gabegorelick:
The loading of config was previously untested.

# Issue Type

- Feature Pull Request

## Summary

Move `parseCommandLineFlags` into `config.go` and add some tests for it.

   to it.
   and
   configurations.


autospotting.go
```diff
@@ -34,43 +30,7 @@ func main() {
...
-		conf.TagFilteringMode,
-		conf.SpotProductDescription,
-		conf.InstanceTerminationMethod,
-		conf.TerminationNotificationAction,
-		conf.CronSchedule,
-		conf.CronScheduleState,
-		conf.LicenseType,
-		conf.PatchBeanstalkUserdata,
-	)
+	log.Printf("Configuration flags: %#v", conf)
```
gabegorelick: Pulled this change from #354.

core/config.go
```diff
@@ -88,3 +93,103 @@ type Config struct {
...
+		"inside or outside the schedule defined by cron_schedule. Allowed values: on|off\n"+
+		"\tExample: ./AutoSpotting --cron_schedule_state='off' --cron_schedule '9-18 1-5'  # would only take action outside the defined schedule\n")
+	flagSet.StringVar(&conf.LicenseType, "license", "evaluation", "\n\tControls the terms under which you use AutoSpotting"+
+		"Allowed values: evaluation|I_am_supporting_it_on_Patreon|I_contributed_to_development_within_the_last_year|I_built_it_from_source_code\n"+
+		"\tExample: ./AutoSpotting --license evaluation\n")
+	flagSet.StringVar(&conf.PatchBeanstalkUserdata, "patch_beanstalk_userdata", "", "\n\tControls whether AutoSpotting patches Elastic Beanstalk UserData scripts to use the instance role when calling CloudFormation helpers instead of the standard CloudFormation authentication method\n"+
+		"\tExample: ./AutoSpotting --patch_beanstalk_userdata true\n")
+
+	printVersion := flagSet.Bool("version", false, "Print version number and exit.\n")
+	flagSet.Parse(os.Args[1:])
```
golangcibot: Error return value of `flagSet.Parse` is not checked (from `errcheck`)

core/config.go
```diff
@@ -88,3 +93,103 @@ type Config struct {
 	// authentication method
 	PatchBeanstalkUserdata string
 }
+
+// ParseConfig loads configuration from command line flags, environments variables, and config files.
+func ParseConfig(conf *Config) {
```
gabegorelick: Renamed from `parseCommandLineFlags` since it loads config from environment variables and config files too.

core/config_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+package autospotting
+
+import (
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
```
gabegorelick: I can remove this, but it makes writing assertions a lot easier. It also has a [package for testing binaries](https://godoc.org/gotest.tools/icmd), which I think will be useful to get `autospotting.go` under test.

core/config.go
gabegorelick: Fixed

core/config_test.go
cristim: Thanks, I'll check it out.

core/config.go
```diff
@@ -88,3 +93,106 @@ type Config struct {
...
+		region = r
+	} else {
+		region = endpoints.UsEast1RegionID
+	}
+
+	conf.LogFile = os.Stdout
+	conf.LogFlag = log.Ldate | log.Ltime | log.Lshortfile
+	conf.MainRegion = region
+	conf.SleepMultiplier = 1
+	conf.LicenseType = os.Getenv("LICENSE")
```
cristim: This looks unnecessary, it should be handled below

gabegorelick: Thanks, that was copied from the old `parseCommandLineFlags`. Updated.

PR #415, opened 2020-02-04
Move logs about incompatible instance types to debug

gabegorelick:
Previously, the amount of logs generated when replacing a large
number of on-demand instances could be overwhelming since info about
each candidate instance type was logged. Now, we still log when an
instance type matches, but now all the logs about every instance type
that does not match will be at debug level.

This change was begun in https://github.com/AutoSpotting/AutoSpotting/pull/354/files#diff-fa23d3c4fb090da3bc9f71371d6e1d95R446.

# Issue Type

- Feature Pull Request

## Summary

Reduce amount of logs generated when attempting to find compatible instance types.

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling_configuration.go
+++ b/core/autoscaling_configuration.go
@@ -243,7 +243,7 @@ func (a *autoScalingGroup) loadConfSpot() bool {
 	}
 	if newValue, done := a.loadBiddingPolicy(tagValue); done {
 		a.region.conf.BiddingPolicy = newValue
-		logger.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
+		debug.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
 		return done
 	}
 	return false
--- a/core/instance.go
+++ b/core/instance.go
@@ -212,7 +212,7 @@ func (i *instance) terminate() error {
 
 func (i *instance) isPriceCompatible(spotPrice float64) bool {
 	if spotPrice == 0 {
-		logger.Printf("\tUnavailable in this Availability Zone")
+		debug.Printf("\tUnavailable in this Availability Zone")
 		return false
 	}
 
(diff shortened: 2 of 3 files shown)


core/autoscaling_configuration.go
```diff
@@ -243,7 +243,7 @@ func (a *autoScalingGroup) loadConfSpot() bool {
 	}
 	if newValue, done := a.loadBiddingPolicy(tagValue); done {
 		a.region.conf.BiddingPolicy = newValue
-		logger.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
+		debug.Println("BiddingPolicy =", a.region.conf.BiddingPolicy)
```
gabegorelick: I don't think this adds anything since the default bidding policy was already logged on startup.

core/instance.go
```diff
@@ -460,16 +460,16 @@ func (i *instance) launchSpotReplacement() error {
 func (i *instance) getPricetoBid(
 	baseOnDemandPrice float64, currentSpotPrice float64) float64 {
 
-	logger.Println("BiddingPolicy: ", i.region.conf.BiddingPolicy)
+	debug.Println("BiddingPolicy: ", i.region.conf.BiddingPolicy)
```
gabegorelick: The 2 logs statements below are enough to infer the bidding policy.

core/instance.go
```diff
@@ -391,7 +391,7 @@ func (i *instance) getCompatibleSpotInstanceTypesListSortedAscendingByPrice(allo
 			i.isStorageCompatible(candidate, attachedVolumesNumber) &&
 			i.isVirtualizationCompatible(candidate.virtualizationTypes) {
 			acceptableInstanceTypes = append(acceptableInstanceTypes, acceptableInstance{candidate, candidatePrice})
-			logger.Println("\tMATCH FOUND, added", candidate.instanceType, "to launch candiates list")
+			logger.Println("\tMATCH FOUND, added", candidate.instanceType, "to launch candiates list for instance", i.InstanceId)
```
gabegorelick: Adding instanceID to some of these logs to help provide context.

gabegorelick:
Thanks for merging this so quickly @cristim.

cristim:
Thanks for your contribution!

I'm doing my best to keep up with you and get these things merged and delivered to the users timely.

PR #417, opened 2020-02-05
Add tools to go.mod

gabegorelick:
This uses the pattern outlined in https://github.com/go-modules-by-example/index/tree/master/010_tools to version tools.

It also fails the build if `go.mod` or `go.sum` are out of date, which ensures that changes to those files are always checked in.

# Issue Type

- Feature Pull Request

   to it.
   and
   configurations.


go.sum
```diff
@@ -1,15 +1,7 @@
-github.com/aws/aws-lambda-go v1.10.0 h1:uafgdfYGQD0UeT7d2uKdyWW8j/ZYRifRPIdmeqLzLCk=
-github.com/aws/aws-lambda-go v1.10.0/go.mod h1:zUsUQhAUjYzR8AuduJPCfhBuKWUaDbQiPOG+ouzmE1A=
```
gabegorelick: Deletions are from `go mod tidy`.

gabegorelick:
@cristim This is finally ready for review.

Makefile
```diff
@@ -34,18 +34,16 @@ check_deps:                                                  ## Verify the syste
 	@echo "all dependencies satisifed: $(DEPS)"
 .PHONY: check_deps
 
-build_deps:
-	@command -v goveralls || go get github.com/mattn/goveralls
-	@command -v golint || go get golang.org/x/lint/golint
-	@go tool cover -V || go get golang.org/x/tools/cmd/cover
+build_deps:                                                  ## Install all dependencies specified in tools.go
+	@grep _ tools.go | cut -d '"' -f 2 | xargs go install
```
gabegorelick: I cannot believe there's no better way to tell go to install all modules specified in go.mod...

tools.go
```diff
@@ -0,0 +1,9 @@
+// +build tools
+
+package tools
+
+import (
+	_ "github.com/mattn/goveralls"
```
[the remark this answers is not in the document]
gabegorelick: Fixed

Makefile
cristim: Doesn't this require network connectivity on every run?

gabegorelick: No, it uses the module cache. The first may download modules, but subsequent invocations do not.

PR #418, opened 2020-02-10
Remove incorrect Makefile conditionals

gabegorelick:
`ifeq` and similar Makefile constructs run _before_ target invocation.
This means that `go vet` and `gofmt` were being run when targets
like `make help` were run.

FWIW, I don't think the extra code added much over simply invoking
`go vet` and `gofmt`. Their output is already very good.

# Issue Type

- Feature Pull Request

## Summary

Don't run `go vet` and `go fmt` in every invocation of `make`.

   to it.
   and
   configurations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,26 +70,12 @@ upload: archive                                              ## Upload binary
 	aws s3 sync build/s3/ s3://$(BUCKET_NAME)/
 .PHONY: upload
 
-vet-check: build_deps                                        ## Verify vet compliance
-ifeq ($(shell go vet -all . | wc -l | tr -d '[:space:]'), 0)
-	@printf "ok\tall files passed go vet\n"
-else
-	@printf "error\tsome files did not pass go vet\n"
-	@go vet -all . 2>&1
-	@exit 1
-endif
-
+vet-check:                                                   ## Verify vet compliance
+	@go vet -all ./...
 .PHONY: vet-check
 
-fmt-check: build_deps                                        ## Verify fmt compliance
-ifeq ($(shell gofmt -l -s . | wc -l | tr -d '[:space:]'), 0)
-	@printf "ok\tall files passed go fmt\n"
-else
-	@printf "error\tsome files did not pass go fmt, fix the following formatting diff:\n"
+fmt-check:                                                   ## Verify fmt compliance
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -70,26 +70,12 @@ upload: archive                                              ## Upload binary
 	aws s3 sync build/s3/ s3://$(BUCKET_NAME)/
 .PHONY: upload
 
-vet-check: build_deps                                        ## Verify vet compliance
-ifeq ($(shell go vet -all . | wc -l | tr -d '[:space:]'), 0)
-	@printf "ok\tall files passed go vet\n"
-else
-	@printf "error\tsome files did not pass go vet\n"
-	@go vet -all . 2>&1
```
gabegorelick: We were running `go vet` twice.

Makefile
```diff
@@ -70,26 +70,12 @@ upload: archive                                              ## Upload binary
...
-	@exit 1
-endif
-
+vet-check:                                                   ## Verify vet compliance
+	@go vet -all ./...
```
gabegorelick: Default output of `go vet` seems fine to me. No need to print extra error messages.

image

Makefile
```diff
@@ -70,26 +70,12 @@ upload: archive                                              ## Upload binary
...
-ifeq ($(shell gofmt -l -s . | wc -l | tr -d '[:space:]'), 0)
-	@printf "ok\tall files passed go fmt\n"
-else
-	@printf "error\tsome files did not pass go fmt, fix the following formatting diff:\n"
-	@gofmt -l -s -d .
-	@exit 1
-endif
-
+fmt-check:                                                   ## Verify fmt compliance
+	@gofmt -l -s -d ./...
```
gabegorelick: Example output for failing gofmt:
image

Makefile
```diff
@@ -70,26 +70,12 @@ upload: archive                                              ## Upload binary
 	aws s3 sync build/s3/ s3://$(BUCKET_NAME)/
 .PHONY: upload
 
-vet-check: build_deps                                        ## Verify vet compliance
```
gabegorelick: Dependency on `build_deps` isn't needed. `go vet` is built in to go AFAIK.

Issue #406, opened 2020-01-30
Move code out of autospotting.go


gabegorelick:
# Github issue #

## Issue type ##

- Feature Idea

## Summary ##

There's currently no good way to (unit) test code in the file `autospotting.go` (please correct me if I'm wrong!) since that's in the `main` package. We should move as much functionality as we can out of this file so that all the business logic can be tested by unit tests.

This is especially important with changes like #354 changing a lot of code in `autospotting.go` but lacking test coverage.

cristim:
I'm not sure this is still relevant, the code was pretty much rewritten since this issue was raised.

autospotting.go now just defines the Lambda handler and some initialization code. Not sure we need unit tests for that.

PR #421, opened 2020-02-14
Move almost everything out of autospotting.go

gabegorelick:
This moves most code from autospotting.go to main.go and adds more tests for the latter.

See #406

# Issue Type

- Feature Pull Request

## Summary

Move code out of the `main` package and add some tests. 

   to it.
   and
   configurations.

### DIFF
--- a/autospotting.go
+++ b/autospotting.go
@@ -4,91 +4,24 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
 	"log"
-	"os"
 
 	autospotting "github.com/AutoSpotting/AutoSpotting/core"
-	"github.com/aws/aws-lambda-go/events"
-	"github.com/aws/aws-lambda-go/lambda"
 )
-
-var conf autospotting.Config
 
 // Version represents the build version being used
 var Version = "number missing"
 
 func main() {
-	if os.Getenv("AWS_LAMBDA_FUNCTION_NAME") != "" {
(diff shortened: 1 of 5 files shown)


gabegorelick:
@cristim My hope was to get as much moved out from `autospotting.go` and then get everything covered by unit tests in preparation for landing larger changes, like the event-based instance replacement, more safely. But there's still a lot of untested code so I'm not sure if this actually helps or just introduces more chances for regressions.

I'm open to suggestions for how to increase test coverage further. I see some old discussion on https://github.com/AutoSpotting/AutoSpotting/issues/172, but curious to hear your current thoughts.

cristim:
Thanks for moving this forward.

I think at some point after merging a number of harmless/easy hunks we'll end up with a large chunk that we'll have to merge at once.

I think I prefer to have a larger jump with regressions that we then address quickly instead of an extended period of maybe more subtle regressions.

cristim:
@gabegorelick I just merged the event-based PR a few days ago, would this PR still be relevant if we can rebase it on top of the current code or can we close it?

cristim:
@gabegorelick are you still interested in contributing this?

There were lots of code changes since this was submitted so perhaps some of this doesn't make sense anymore.

gabegorelick:
> are you still interested in contributing this?

Sorry, I probably won't have time.

PR #424, opened 2020-03-27
Update how bid price is calculated for premium instances

tapirs:
# Issue Type

- Bugfix Pull Request

## Summary

A bug fix for PR #423, currently when calculating the bid price for premium instances, using the aggressive policy, the premium is included when calculating the buffer and as such this adds on a buffer for the premium which should be fixed.

For example

A standard instance

ondemand price = 0.1
spot price = 0.05
premium = 0.0
buffer = 50%

would give a bid price of 0.075, which would be lower than the on demand price.

A premium instance

ondemand price = 0.16
spot price = 0.11
premium = 0.06
buffer = 50%

would give a bid price of 0.165, which would be higher than the on demand price.

A premium instance after this fix

ondemand price = 0.16
spot price = 0.11
premium = 0.06
buffer = 50%

would give a bid price of 0.135, which would be lower than the on demand price.

   to it.
   and
   configurations.

### DIFF
--- a/core/instance.go
+++ b/core/instance.go
@@ -427,7 +427,7 @@ func (i *instance) launchSpotReplacement() error {
 	for _, instanceType := range instanceTypes {
 		az := *i.Placement.AvailabilityZone
 		bidPrice := i.getPricetoBid(i.price,
-			instanceType.pricing.spot[az])
+			instanceType.pricing.spot[az], instanceType.pricing.premium)
 
 		runInstancesInput := i.createRunInstancesInput(instanceType.instanceType, bidPrice)
 		logger.Println(az, i.asg.name, "Launching spot instance of type", instanceType.instanceType, "with bid price", bidPrice)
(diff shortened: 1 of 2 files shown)


cristim:
The premium also applies to the on demand price, right?

Please correct me if I'm wrong but shouldn't we be comparing with 0.22, which is the on demand + premium price, so 0.165 is still below this on demand premium price.

tapirs:
The calculations above already have the premium included for the on-demand price.

At the point in the code, the on-demand price will include the premium so we'll be comparing the on-demand price and the base spot price plus the premium. 

So the last 2 examples again

A premium instance

ondemand price = 0.16 (linux price + premium => 0.1 + 0.06)
spot price = 0.11 (linux spot price + premium => 0.05 +0.06)
premium = 0.06
buffer = 50%

would give a bid price of 0.165 (0.11 * 1.5), which would be higher than the on-demand price of 0.16.

A premium instance after this fix

ondemand price = 0.16 (linux price + premium => 0.1 + 0.06)
spot price = 0.11 (linux spot price + premium => 0.05 +0.06)
premium = 0.06
buffer = 50%

would give a bid price of 0.135 (((0.11 - 0.06) * 1.5) + 0.06), which would be lower than the on-demand price of 0.16.

I noticed this because any instance smaller than a t3.medium were still not being picked up when using RHEL and because the margins are so narrow the 10% extra from the premium was enough to make the spot price higher than the on-demand price. I've applied this fix to my environments and it is now replacing t2.small and t2.micro instance.

cristim:
Makes sense now, thanks!

PR #425, opened 2020-04-02
Feat/event based instance replacement - Fix to terminateRandomSpotInstance logic in Schedule run + specify ondemandprice multiplier on a ASG group level

mello7tre:
# Issue Type

- Bugfix Pull Request
- Feature Pull Request

### Bugfix 
i found a problem in terminateRandomSpotInstanceIfHavingEnough logic that cause a loop of termination and running instance.
Consider the following setup:
ASG Min = Max = Desired = 2
ASG configured for Autospotting MinOnDemand = 1

Activate Autospotting, first only scheduled run.
After some time you will arrive at the following state:
1 Running onDemand
1 Running Spot.

Next time autspotting wil be executed it will enter a loop of termination and running.

Lets consider first cronEventAction, i will post only relevant lines:
```
spotInstance := a.findUnattachedInstanceLaunchedForThisASG()
 if spotInstance == nil {
        onDemandInstance := a.getAnyUnprotectedOnDemandInstance()
        if onDemandInstance == nil {
            return enableEventHandling{target{asg: a}}
         }
         if need, total := a.needReplaceOnDemandInstances(); !need {
           return terminateSpotInstance{target{asg: a, totalInstances: total}}
         }
}
```
as spotInstance is nil and onDemandInstance != nill we will call needReplaceOnDemandInstances, this will return false so we will end in:
terminateSpotInstance

this in return will call terminateRandomSpotInstanceIfHavingEnough

let's consider it (relevant lines too):
(totalRunning is total from previous methods)
```
func (a *autoScalingGroup) terminateRandomSpotInstanceIfHavingEnough(totalRunning int64, wait bool) error {
        if totalRunning == 1 {
                return nil
        }
        if a.allInstancesRunning() && a.instances.count64() < *a.DesiredCapacity {
                logger.Println("Not enough capacity in the group")
                return nil
        }
-------> after this point a randomly-selected spot instance is terminated
        randomSpot := a.getAnySpotInstance()
        if randomSpot == nil {
                return nil
        }
        logger.Println("Terminating randomly-selected spot instance",
                *randomSpot.Instance.InstanceId)
}
```
In our example we have a.instances.count64() = 2 and a.DesiredCapacity = 2 too, so even if all instances are running (a.allInstancesRunning() return true) we do not enter in second condition and we go on to terminating a random spot instance.
So in our example the only spot instance is terminated, ASG launch a new one, after some autopsotting cycle we return to our starting point (1 ondemand and 1 spot) and we re-enter the loop.
(loop is even faster if we have event-based replacement enabled)

This PR fix it.

### Feature
OnDemand Price Multiplier can now be specified at ASG level using the tag:
**autospotting_on_demand_price_multiplier**

This way, choosing a proper value, is possible to have a single instance of Autospotting to manage both ASG Linux and ASG Windows.
(usually ondemand price for a windows instance is ~ 1.8 greater than a linux one). 

   to it.
   and
1. [ X] Tags names and expected values should be similar to the other existing
   configurations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine as golang
+FROM golang:1.13-alpine as golang
 RUN apk add -U --no-cache ca-certificates git make
 COPY . /src
 WORKDIR /src
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine
+FROM golang:1.13-alpine
 RUN apk add -U --no-cache ca-certificates git make zip
 
 COPY . /src
--- a/START.md
+++ b/START.md
@@ -189,10 +189,53 @@ be risky, please handle with care.
 
 ### For Elastic Beanstalk ###
 
-* In order to add tags to existing Elastic Beanstalk environment, you will need
-  to rebuild or update the environment with the `spot-enabled` tag. For more
-  details you can follow this
-  [guide](http://www.boringgeek.com/add-or-update-tags-on-existing-elastic-beanstalk-environments)
+Elastic Beanstalk uses CloudFormation to create an Auto-Scaling Group. The ASG
+is then in charge of automatically scaling your application up and down. As a
+result, AutoSpotting works natively with Elastic Beanstalk.
+
+Follow these steps to configure AutoSpotting with Elastic Beanstalk.
+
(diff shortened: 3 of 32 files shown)


mello7tre:
i have fixed the tests that was failing after the work on this branch.
Now go test terminate without error.
I have added a simply test for my fix to terminateRandomSpotInstanceIfHavingEnough  logic.

PR #426, opened 2020-04-08
Cron timezone

tapirs:
Added the ability to select a timezone to check if autospotting should be run based on the users own timezone

# Issue Type

- Feature Pull Request

## Summary

Sets the timezone in which to check the CronSchedule. Also takes into account daylight savings in the timezone.

Example: If the timezone is set to 'UTC' and the CronSchedule is '9-18 1-5' it would start the interval at 9am UTC, with the timezone set to 'Europe/London' it would start the interval at 9am BST (10am UTC) or 9am GMT (9am UTC) depending on daylight savings.

   to it.
   and
   configurations.


cristim:
@tapirs I had a look into this and looks like we need to update the dependencies in order to fix the build.

I'll do it on master then we can rebase this Pull Request

tapirs:
@cristim The build is now passing.

tapirs:
@cristim Done :)

PR #436, opened 2020-08-25
Automatically upgrade EBS volumes of launched spot instances from GP2->GP3 and IO1->IO2 for lower cost and/or higher performance

cristim:
# Issue Type

- Feature Pull Request

## Summary

As per https://aws.amazon.com/blogs/aws/new-ebs-volume-type-io2-more-iops-gib-higher-durability/ they have 100x Higher Durability and 10x More IOPS/GiB at the same cost. Similar awesomeness is available on GP3 as per https://aws.amazon.com/blogs/aws/new-amazon-ebs-gp3-volume-lets-you-provision-performance-separate-from-capacity-and-offers-20-lower-price/

What we have so far:
    - Added configuration support with per-ASG tag-based overrides for the GP2 threshold
    - Set default threshold value to 170GB, apparently that's the threshold where GP2 bandwidth overtakes the GP3, but for IOPS the threshold is at 1TB. This is configurable at the Lambda binary level but not yet exposed through CloudFormation/Terraform
    - Add regional blacklist for the IO1->IO2 conversion for the regions where it's not yet supported
    - Reworked LaunchConfiguration and LaunchTemplate processing when creating the RunInstances input data
    - Added support for also processing the Block Device Mappings set by the AMI, with new overrides from the Launch Template level in addition to the existing ones for Launch Configurations
    - Added basic unit tests for the Launch Template conversion process
    - Tested and working as expected using ASG tag overrides for instances having BDMs configured at the AMI level, but without LC overrides.
    
ToDo:
    - Add CloudFormation and Terraform support for setting the Lambda environment variables based on stack parameter for overriding the 170GB default GP2 Threshold value.
    - Test the LC and LT overrides
    - Add tests for the AMI and LC BDM conversion logic

Current test results, with threshold set at 20 GB using ASG tag overrides:
image

   to it.
1.  [x] No issues are reported when running `make full-test`.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -739,29 +739,6 @@ func (a *autoScalingGroup) terminateInstanceInAutoScalingGroup(
 	return nil
 }
 
-func (a *autoScalingGroup) hasLaunchLifecycleHooks() (bool, error) {
-
-	resDLH, err := a.region.services.autoScaling.DescribeLifecycleHooks(
-		&autoscaling.DescribeLifecycleHooksInput{
-			AutoScalingGroupName: a.AutoScalingGroupName,
-		})
-
-	if err != nil {
-		log.Println(err.Error())
-		return false, err
-	}
-
-	for _, hook := range resDLH.LifecycleHooks {
-		if *hook.LifecycleTransition == "autoscaling:EC2_INSTANCE_LAUNCHING" {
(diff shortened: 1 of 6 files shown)


cristim:
I'll just merge it as it is now and iterate on it further based on user feedback

PR #445, opened 2020-10-22
Add log output after having processed all regions that show actions taken by AS ScheduleRun

mello7tre:
# Issue Type

- Feature Pull Request

## Summary
As sometimes it's difficult to parse log output to find out which action have been taken by AutoSpotting schedule run, especially if you have many ASG, this PR add a final output after having processed all regions that show only if some action have been taken on an AutoScalingGroup (instance terminated, replaced or spot launched and so on)
Here an example output:
```
           SC:2020-10-22T11:40:00 2020/10/22 11:41:12 main.go:74: ####### BEGIN FINAL RECAP #######
           SC:2020-10-22T11:40:00 2020/10/22 11:41:12 main.go:77: eu-west-1 rmq-u-d-AutoScalingGroup-E0AA2BJJLA1I OnDemand instance i-02feae0a90d03651d replaced with spot instance i-04caa6efea8a30242
           SC:2020-10-22T11:16:00 2020/10/22 11:16:40 main.go:77: eu-west-1 els-u-d-AutoScalingGroup-V4UV43APWMHZ Launched spot instance i-08791721f770db501
           SC:2020-10-22T11:22:00 2020/10/22 11:22:38 main.go:77: eu-central-1 ecs-u-d-AutoScalingGroup-16M9XBTDZ43M2 Terminated spot instance i-00cca1a2835ebf303 [not needed]
           2020/10/22 11:41:12 Execution completed, nothing left to do
           END RequestId: e835f04a-a58a-421a-b5ea-66e1dc02912a
           REPORT RequestId: e835f04a-a58a-421a-b5ea-66e1dc02912a       Duration: 34836.55 ms   Billed Duration: 34900 ms       Memory Size: 256 MB     Max Memory Used: 178 MB
```
To achieve this i added a new property to Config called FinalRecap as a map of string slice.
As Config is passed as a pointer and as region name is the map key, there should be no problem regarding concurrent/parallel region execution changing the same "property".

Note:
I had to clear every time FinalRecap map (in ProcessCronEvent) because config may persist over differents AutoSpotting runs.

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -204,7 +204,9 @@ func (a *autoScalingGroup) cronEventAction() runer {
 	logger.Println("Found unattached spot instance", spotInstanceID)
 
 	if need, total := a.needReplaceOnDemandInstances(); !need || !shouldRun {
-
+		// add to FinalRecap
+		recapText := fmt.Sprintf("%s %s Terminated not needed instance %s", a.region.name, a.name, spotInstanceID)
+		a.region.conf.FinalRecap[a.region.name] = append(a.region.conf.FinalRecap[a.region.name], recapText)
 		return terminateUnneededSpotInstance{
 			target{
 				asg:            a,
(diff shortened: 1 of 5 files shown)


cristim (approved):
Looks great, thanks!

PR #446, opened 2020-10-31
re #430: allows counting ephemeral volumes desired via AMI specified by ASG launch template (w/ version)

### LINKED ISSUE #430: autospotting not respecting presence of ephemeral volume in AMI/launch template
# Github issue #

## Issue type ##

- Bug Report

## Build number #

```
nightly-1137
```
believe we used master branch code from end of April 2020.

## Environment ##

- AWS region: us-east-1
- Type of environment: VPC

## Summary ##

I've got an ASG set up to launch instances from a Launch Template using m5d.xlarge instances. However autospotting is allowing spot instances without an ephemeral volume to be used.

I've attempted to detail my setup below, however if there's some config I've missed that makes the ephemeral volume requirement apparent to autospotting in a different way than I attempted please do let me know! :)

## Steps to reproduce ##

The AMI used for this template requests an ephemeral volume:
```
Block Devices
/dev/xvda=snap-0a74061116e945b70:16:true:gp2, /dev/sdb=ephemeral0
```

The launch template says: `Currently no volumes details are specified and therefore the instances launch with the template default volume settings`. When I go to modify the template I can only add EBS volumes, but the instance store volume is listed:

Screen Shot 2020-05-13 at 11 40 40 AM

Debug output from autospotting shows:
```
2020/05/13 12:54:43 instance.go:373: Comparing current type m5d.xlarge with price 0.226 with candidate m5.xlarge with price 0.0815
2020/05/13 12:54:43 instance.go:327: Checking allowed/disallowed list
2020/05/13 12:54:43 instance.go:219: Comparing class spot/instance:
2020/05/13 12:54:43 instance.go:220: 	Spot CPU/memory/GPU:  4  /  16  /  0
2020/05/13 12:54:43 instance.go:222: 	Instance CPU/memory/GPU:  4  /  16  /  0
2020/05/13 12:54:43 instance.go:286: Comparing storage spot/instance:
2020/05/13 12:54:43 instance.go:287: 	Spot volumes/size/ssd:  0 0 false <-----------
2020/05/13 12:54:43 instance.go:291: 	Instance volumes/size/ssd:  0 150 true <-----------
2020/05/13 12:54:43 instance.go:312: Comparing virtualization spot/instance:
2020/05/13 12:54:43 instance.go:313: 	Spot virtualization:  [HVM]
2020/05/13 12:54:43 instance.go:314: 	Instance virtualization:  hvm
2020/05/13 12:54:43 instance.go:383: 	MATCH FOUND, added m5.xlarge to launch candiates list
```

I took a quick look through the code and came upon:

```
	usedMappings := i.asg.launchConfiguration.countLaunchConfigEphemeralVolumes()
	attachedVolumesNumber := min(usedMappings, current.instanceStoreDeviceCount)
...
	i.isStorageCompatible(candidate, attachedVolumesNumber) &&
```

Stands to reason one of those terms is 0 and the min is going with that. I suspect it's usedMappings but not entirely sure.

## Expected results ##

Expecting launch template with AMI having ephemeral volume specified to require instance types that have an ephemeral volume available.

## Actual results ##

```text
```

karock:
# Issue Type

- Feature Pull Request

## Summary

Fixes #430, where ephemeral volumes weren't being counted from the AMI used by a launch template when considering the acceptable instance types.

   to it.
   and
   configurations.

### DIFF
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -96,6 +96,10 @@ func (a *autoScalingGroup) loadLaunchTemplate() (*launchTemplate, error) {
 		return nil, err
 	}
 
+	if len(resp.LaunchTemplateVersions) == 0 {
+		return nil, errors.New("missing launch template")
+	}
+
 	ltv := resp.LaunchTemplateVersions[0]
 
 	params2 := &ec2.DescribeImagesInput{
@@ -107,6 +111,10 @@ func (a *autoScalingGroup) loadLaunchTemplate() (*launchTemplate, error) {
 	if err2 != nil {
 		logger.Println(err.Error())
 		return nil, err
+	}
+
+	if len(resp2.Images) == 0 {
+		return nil, errors.New("missing launch template image")
 	}
 
 	a.launchTemplate = &launchTemplate{
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -69,8 +69,14 @@ func (a *autoScalingGroup) loadLaunchTemplate() (*launchTemplate, error) {
 		return a.launchTemplate, nil
 	}
 
-	ltID := a.LaunchTemplate.LaunchTemplateId
-	ltVer := a.LaunchTemplate.Version
+	lt := a.LaunchTemplate
+
+	if lt == nil {
+		return nil, errors.New("missing launch template")
+	}
+
+	ltID := lt.LaunchTemplateId
+	ltVer := lt.Version
 
 	if ltID == nil || ltVer == nil {
 		return nil, errors.New("missing launch template")
--- a/core/autoscaling.go
+++ b/core/autoscaling.go
@@ -27,6 +27,7 @@ type autoScalingGroup struct {
 	name                string
 	region              *region
 	launchConfiguration *launchConfiguration
+	launchTemplate      *launchTemplate
 	instances           instances
 	minOnDemand         int64
 	config              AutoScalingConfig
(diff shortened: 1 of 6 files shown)


core/autoscaling.go
```diff
@@ -62,6 +63,59 @@ func (a *autoScalingGroup) loadLaunchConfiguration() (*launchConfiguration, erro
...
+	}
+
+	resp, err := svc.DescribeLaunchTemplateVersions(params)
+
+	if err != nil {
+		logger.Println(err.Error())
+		return nil, err
+	}
+
+	ltv := resp.LaunchTemplateVersions[0]
```
cristim, line 99: I would have a check above in case the length of resp.LaunchTemplateVersions is zero

core/autoscaling.go
```diff
@@ -62,6 +63,59 @@ func (a *autoScalingGroup) loadLaunchConfiguration() (*launchConfiguration, erro
...
+
+	params2 := &ec2.DescribeImagesInput{
+		ImageIds: []*string{ltv.LaunchTemplateData.ImageId},
+	}
+
+	resp2, err2 := svc.DescribeImages(params2)
+
+	if err2 != nil {
```
cristim, line 107: also add a condition here for the length of resp.Images, which below at line 114 we assume to be at least 1

core/autoscaling.go
```diff
@@ -205,6 +259,7 @@ func (a *autoScalingGroup) cronEventAction() runer {
 		}
 
 		a.loadLaunchConfiguration()
+		a.loadLaunchTemplate()
```
cristim, line 262: why do we return anything on `a.loadLaunchTemplate` if we ignore both its return values?

core/instance.go
```diff
@@ -917,3 +920,9 @@ func min(x, y int) int {
 	}
 	return y
 }
+func max(x, y int) int {
```
cristim, line 923: do we need to define this again? I think we had it defined somewhere else.

Hopefully we'll soon get some sort of generics support in Go to avoid the need to define this sort of things.

core/autoscaling.go
karock: I was using loadLaunchConfiguration as my template, looks like it has the same issue. In both cases only the tests use the return values. I can change it if you like though.

core/instance.go
karock: I only saw `min` defined in this way, right above where I added this change.

cristim: Okay, thanks

core/autoscaling.go
karock: addressed in https://github.com/AutoSpotting/AutoSpotting/pull/446/commits/c1153e58d76ad5d84dd1e5a30a0ddfaa823f26c0

core/autoscaling.go
karock: addressed in https://github.com/AutoSpotting/AutoSpotting/pull/446/commits/c1153e58d76ad5d84dd1e5a30a0ddfaa823f26c0

cristim:
thanks for your contribution @karock

PR #451, opened 2020-11-11
Add explicit bucket permissions for regional resources deployment

gjmveloso:
# Issue Type

- Bugfix Pull Request

## Summary

When trying to launch AutoSpotting in my AWS account I got the following errors during creation of regional stacks (StackSet disabled)

>
	2020-11-11T10:22:53.462-03:00	Traceback (most recent call last):
	2020-11-11T10:22:53.462-03:00	File "/var/lang/lib/python3.6/threading.py", line 916, in _bootstrap_inner
	2020-11-11T10:22:53.462-03:00	self.run()
	2020-11-11T10:22:53.462-03:00	File "/var/lang/lib/python3.6/threading.py", line 864, in run
	2020-11-11T10:22:53.462-03:00	self._target(*self._args, **self._kwargs)
	2020-11-11T10:22:53.462-03:00	File "/var/task/regional_stack_lambda.py", line 47, in create_stack
	2020-11-11T10:22:53.462-03:00	'ParameterValue': role_arn,
	2020-11-11T10:22:53.462-03:00	File "/var/runtime/botocore/client.py", line 316, in _api_call
	2020-11-11T10:22:53.462-03:00	return self._make_api_call(operation_name, kwargs)
	2020-11-11T10:22:53.462-03:00	File "/var/runtime/botocore/client.py", line 635, in _make_api_call
	2020-11-11T10:22:53.462-03:00	raise error_class(parsed_response, operation_name)
	2020-11-11T10:22:53.462-03:00	botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateStack operation: S3 error: Access Denied
	2020-11-11T10:22:53.462-03:00	For more information check http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html 

   to it.
   and
   configurations.

### DIFF
--- a/cloudformation/stacks/AutoSpotting/template.yaml
+++ b/cloudformation/stacks/AutoSpotting/template.yaml
@@ -864,6 +864,13 @@
                 - "lambda:GetFunctionConfiguration"
               Effect: "Allow"
               Resource: "*"
+            - 
+              Action:
+                - "s3:Get*"
+              Effect: "Allow"
+              Resource: 
+                - Fn::Sub: "arn:${AWS::Partition}:s3:::${LambdaS3Bucket}"
+                - Fn::Sub: "arn:${AWS::Partition}:s3:::${LambdaS3Bucket}/${LambdaS3BucketPrefix}/*"
         PolicyName: "LambdaRegionalStackPolicy"
         Roles:
           -


cristim:
Thanks for your contribution, but I think this should be obsoleted by #449.

gjmveloso:
That's true but StackSets are still in progress and an option in the main Cfn template -- with default parameter set to not use StackSets.

Fixing this allow new users to spin AutoSpotting with default parameters without issues.

cristim:
I'm going to merge it, thanks!

The StackSet parameter in the current Cloudformation template is for using the same template but leveraging a multi-regional StackSet to deploy different resources across regions based on the heavy use of conditionals, and also significant duplication to the regional Cloudformation template.

When disabled, it is expected to be installed in a single main region and only deploys there a smaller set of resources and uses a Lambda backed custom resource to deploy the second regional Cloudformation template on all regions. This is overcomplicated and hard to maintain, so that's why I wanted to simplify it.

The new and still in progress StackSet pull request simplifies this messy template. It uses different templates, a simplified template for the main region that no longer uses those conditionals but only deploys the main resources and instead of the custom Lambda resource it uses a StackSet to deploy the regional template on all regions. As of today this template can also be deployed across an Organization using a StackSet, and no longer needs the StackSet parameter.

cristim (approved):

PR #458, opened 2021-04-25
New workflow to get rid of LambdaManageASG using SQS FIFO Queue with GroupId equals to ASG Name

mello7tre:
# Issue Type

- Feature Pull Request

## Summary
Based on the idea #457 

Now the workflow for detecting spot launch is (for onDemand nothing is changed):
* Cron schedule:
  * in `replaceOnDemandInstanceWithSpot` we simply check for spot existence and send sqs message to queue
* LifecycleHook trigger:
  * At the end of `handleLifecycleHookEvent` we send a message to the queue
* Spot instance launch event:
  * in `handleNewSpotInstanceLaunch` we send message to sqs queue only if ASG is event enabled.

* SqsQueue lambda trigger
  * message to queue trigger the same code of event based (sqsMessage is a json like the instance launch cloudwatch event) but skip check on `asg.isEnabledForEventBasedInstanceReplacement`
  * spot instance is attached using `swapWithGroupMember`

Another great improvement, apart getting rid of LambdaManageASG, is that now we use only one method/function to replace onDemand with Spot, `swapWithGroupMember`.
Using fifo queue with GroupId equals to ASG name, all replacement operation on the same ASG are sequential and safe regarding changing ASG parameters (max size, suspending proccess ...).

I have done tests for both event and cron onDemand replacement (not LifeCycle, but it should work).
i have AutoSpotting built from this code running in dev environment without any problems.

Code tests have been fixed only to avoid errors not ,at the moment, regarding program logic (Ex. `replaceOnDemandInstanceWithSpot` now simply send a message to the SqsQueue, so some tests need to be changed or are no more necessary)

1. [ X] I hereby allow the Copyright holder the rights to distribute this piece of
   to it.
   and
   configurations.


core/main.go
```diff
@@ -250,7 +254,11 @@ func (a *AutoSpotting) EventHandler(event *json.RawMessage) {
 			log.Println("Couldn't get instance ID of newly launched instance", err.Error())
 			return
 		} else if instanceID != nil {
-			logger.SetPrefix(fmt.Sprintf("ST:%s ", *instanceID))
+			if len(a.config.sqsReceiptHandle) == 0 {
```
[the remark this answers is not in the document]
mello7tre: sorry, have no idea how to simply this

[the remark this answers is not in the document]
cristim: probably we could split this huge and hard to follow branching logic with smaller calls to new functions that take care of only a given sub-branch.

cloudformation/stacks/AutoSpotting/template.yaml
```diff
@@ -376,8 +381,8 @@
               Ref: "TerminationNotificationAction"
             PATCH_BEANSTALK_USERDATA:
               Ref: "PatchBeanstalkUserdata"
-            LAMBDA_MANAGE_ASG:
-              Ref: "LambdaManageASG"
+            SQS_QUEUE_URL:
```
cristim, line 384: Is this queue now compulsory or can AutoSpotting still run in cron mode without this new FIFO queue?

I know of a few people who run it on Kubernetes from the Docker container images, and they probably can't easily provision an SQS queue.

core/config.go
```diff
@@ -121,7 +124,8 @@ func ParseConfig(conf *Config) {
 	conf.LogFlag = log.Ldate | log.Ltime | log.Lshortfile
 	conf.MainRegion = region
 	conf.SleepMultiplier = 1
-	conf.LambdaManageASG = os.Getenv("LAMBDA_MANAGE_ASG")
+	conf.SQSQueueURL = os.Getenv("SQS_QUEUE_URL")
```
cristim, line 127: As far as I know the command like flag library we use shouldn't need this, it's automatically taking env var `VAR_NAME` if we have a flag `-var-name`, and I see we have it defined below.

core/instance.go
```diff
@@ -222,16 +222,16 @@ func (i *instance) terminate() error {
 	return err
 }
 
-func (i *instance) shouldBeReplacedWithSpot() bool {
+func (i *instance) shouldBeReplacedWithSpot(doNotCheckEventEnabled bool) bool {
```
cristim, line 225: why do we need this new `doNotCheckEventEnabled` flag?

core/main.go
```diff
@@ -177,30 +177,34 @@ func (a *AutoSpotting) getRegions() ([]string, error) {
...
 
-	// If the event comes from Sns - extract the Cloudwatch event embedded in it
-	if snsEvent.Records != nil {
-		snsRecord := snsEvent.Records[0]
-		parseEvent = []byte(snsRecord.SNS.Message)
+	// If the event comes from Sqs - extract the Cloudwatch event embedded in it
+	if sqsEvent.Records != nil {
+		sqsRecord := sqsEvent.Records[0]
+		parseEvent = []byte(sqsRecord.Body)
+		a.config.sqsReceiptHandle = sqsRecord.ReceiptHandle
```
cristim, line 197: why do we need to persist this `ReceiptHandle` into the configuration?

core/main.go
```diff
@@ -250,7 +254,11 @@ func (a *AutoSpotting) EventHandler(event *json.RawMessage) {
 			log.Println("Couldn't get instance ID of newly launched instance", err.Error())
 			return
 		} else if instanceID != nil {
-			logger.SetPrefix(fmt.Sprintf("ST:%s ", *instanceID))
+			if len(a.config.sqsReceiptHandle) == 0 {
+				logger.SetPrefix(fmt.Sprintf("ST:%s ", *instanceID))
```
cristim, line 258: what does ST and SQ stand for in the logger prefix?

I'd rather spell entirely whatever that stands for.

core/main.go
```diff
@@ -453,25 +457,18 @@ func (a *AutoSpotting) handleNewSpotInstanceLaunch(r *region, i *instance) error
...
+		}
+		return nil
+	}
+	defer i.region.sqsDeleteMessage(i.InstanceId)
+
 	logger.Printf("%s Found instance %s is not yet attached to its ASG, "+
 		"attempting to swap it against a running on-demand instance",
 		i.region.name, *i.InstanceId)
 
-	hasHooks, err := asg.hasLaunchLifecycleHooks()
```
cristim, line 460: so we no longer handle these launch lifecycle hooks?

core/region.go
```diff
@@ -474,3 +476,54 @@ func (r *region) findEnabledASGByName(name string) *autoScalingGroup {
 	}
 	return nil
 }
+
+func (r *region) sqsSendMessageSpotInstanceLaunch(asgName *string, instanceID *string, instanceState *string) error {
+	inputJSON := "{\"version\":\"0\",\"id\":\"890abcde-f123-4567-890a-bcdef1234567\"," +
```
cristim, line 481: I don't like this defined inline, can we convert it to a Go template?

I'm also unclear on why we even need to do this.

core/region.go
```diff
@@ -474,3 +476,54 @@ func (r *region) findEnabledASGByName(name string) *autoScalingGroup {
...
+	return nil
+}
+
+func (r *region) sqsDeleteMessage(instanceID *string) error {
+	svc := r.services.sqs
+
+	_, err := svc.DeleteMessage(
+		&sqs.DeleteMessageInput{
+			QueueUrl:      &r.conf.SQSQueueURL,
+			ReceiptHandle: &r.conf.sqsReceiptHandle,
```
cristim, line 516: is this the receipt handle we set in the ASG code? to me this looks like a different variable, it might not be the same value.

core/main.go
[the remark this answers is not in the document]
mello7tre: ok, i am working on this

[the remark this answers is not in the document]
mello7tre: made some big changes, need to test them before pushing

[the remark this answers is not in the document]
cristim: thanks Alberto! I really appreciate your great work on this and your previous AutoSpotting improvements

let me know if there's anything I can help you with.

cloudformation/stacks/AutoSpotting/template.yaml
mello7tre: The only way to avoid any concurrency has been to handle all ondemand replacement by the sqs fifo queue.
So now cron events use the queue too.
If we do not do it this way there could be a probability, low but does exist, that two AutoSpotting Lambda invocation, one triggered by sqs queue and one triggered by cron schedule, act in same time on the same ASG.

But people running AutoSpotting in docker images, how are able to invoke LambdaManageASG in the current implementation ?

We can think about a flag, if not using the sqs queue we can in `replaceOnDemandInstanceWithSpot` execute internally swapWithGroupMember instead of sending the message to the sqsqueue.
So to keep only one method for replacing ondemand.
But probably we should need to make some changes to the code.
It's just an idea.

core/main.go
mello7tre: ST is there from the event branch, is standing for "start".
About SQ, as internally sqs trigger, after having "extracted" cloudwatch event from the message, is handled as a spot instance start event, i wanted a way to identify in the log that the event source is sqs.
I am using two uppercase letter to avoid a too longer prefix.
In the current `EventHandler` refactor (not yet pushed) i have changed them to:
```
EC2 Instance State-change Notification -> IS
EC2 Spot Instance Interruption Warning -> II
EC2 Instance Rebalance Recommendation -> IR
AWS API Call via CloudTrail -> CT
Scheduled Event -> SC
```
We can have more spelling prefix, but should be better to have them of the same lenght and not too long.

core/instance.go
mello7tre: i have explained it here:
https://github.com/mello7tre/AutoSpotting/commit/710149391d83ef4e11e7e5536563eb01da4d0676#commitcomment-49924628

As now all replacement are handled by sqsqueue and the method excuted to replace is always `swapWithGroupMember` i need this flag to let replacement work even if ASG is not enabled to process events (only cron schedule).
For spot start event i check that ASG is enabled or not before sending message to sqsqueue.

core/main.go
mello7tre: we need it to delete the message

core/main.go
mello7tre: we handle it by sending a message to the sqsqueue.
But we no more need to check in spot event start if ASG has LifecycleHooks as LifecycleHooks is used when we do not have events enabled

core/region.go
mello7tre: i replaced all ondemand replacement with a call that send a message to the sqsqueue.
The message body is equals to the cloudwatch event of the relative spot instance start.
I just need:
* time, so that sqs content deduplication can work (two message for the same instance are treated as different one)
* instanceID
* instanceState
* region

other field are not used.

core/region.go
mello7tre: yes is the same, and is working correctly.
Sqs send and receive is handled all in the same region of the main Lambda.

cloudformation/stacks/AutoSpotting/template.yaml
cristim: Good point, I haven't seen any bug reports about this but it's probably broken. There are few people who use the Docker images though, so if there are any reports from them we can get this done later, I don't want to overcomplicate this PR as it's already quite complex

core/main.go
cristim: Agreed, but I would prefer to have these documented somewhere or at least in code comments. Probably it wouldn't hurt to have them in-line in the logs, to avoid having to look them up

core/instance_events.go
```diff
@@ -0,0 +1,55 @@
...
+		logger.Println(err.Error())
+		return "", nil, nil, err
+	}
+	eventType := event.DetailType
+
+	// Amazon EC2 State Change Events
+	if eventType == "EC2 Instance State-change Notification" &&
+		detailData.InstanceID != nil &&
+		detailData.State != nil {
+		return "IS", detailData.InstanceID, detailData.State, nil
```
cristim, line 26: Looks much better, I would just like the "IS" and other strings including those of EventBridge events to be defined as constants with descriptive names.

mello7tre: can you give me an example, please

cristim: something like this
```
const InstanceStateNotficationMessage = "EC2 Instance State-change Notification"
const InstanceStateChange = "ISC" // probably better named than "IS"
```
and so on...

The benefit is we get IDE support for them which reduces the chance of errors and increases clarity of the code.

cloudformation/stacks/AutoSpotting/template.yaml
cristim: @mello7tre how hard would it be to make this depend on the presence of the SQS queue in the configuration and if there's no queue defined just handle it without the queue?

mello7tre: i would like to keep a single method `swapWithGroupMember` to replace ondemand, so need to look how to change code to use it directly, if conf.SQSQueueURL is empty, from current cron code.
Currently i am working on https://github.com/AutoSpotting/AutoSpotting/pull/458#discussion_r621955442.
When that part is stable i will give this a look.

core/instance_events.go
```diff
@@ -10,46 +10,73 @@ import (
...
+type instanceData struct {
+	InstanceID     *string `json:"instance-id"`
+	InstanceAction *string `json:"instance-action"`
+	State          *string `json:"state"`
+}
+
 // returns the InstanceID, State or an error
 func parseEventData(event events.CloudWatchEvent) (string, *string, *string, error) {
 	var detailData instanceData
+	var eventTypeCode string = ""
```
cristim, line 46: I don't think we need this initialization, aren't those the Go default values?

core/main.go
```diff
@@ -210,13 +210,13 @@ func (a *AutoSpotting) convertRawEventToCloudwatchEvent(event *json.RawMessage)
...
 func (a *AutoSpotting) processEventInstance(eventType string, region string, instanceID *string, instanceState *string) error {
-	if eventType == "IS" {
+	if eventType == InstanceStateChangeNotificationCode {
 		// If event is Instance state change
 		if len(a.config.sqsReceiptHandle) != 0 {
 			logger.SetPrefix(fmt.Sprintf("SQ:%s ", *instanceID))
 		}
 		a.handleNewInstanceLaunch(region, *instanceID, *instanceState)
-	} else if eventType == "II" || eventType == "IR" {
+	} else if eventType == SpotInstanceInterruptionWarningCode || InstanceRebalanceRecommendationCode == "IR" {
```
cristim, line 219: I haven't tested this but I think this condition might be incorrect

mello7tre: sorry, was a typo ...i fix it

cristim: No problem, this is why we have code reviews 😁

cristim (commented):
Looks great, thanks!

now we just need to make sure the test coverage isn't decreasing. Once that's done it LGTM.

mello7tre:
need  to create a test for parseEventData, before there where only test specific for spot termination/interruption, but now that func do not exist anymore

cristim (approved):

cristim:
Thanks, are we ready to merge this?

mello7tre:
i have it running in dev account for days without any issue.
Every day onDemand instances are replaced with spot (cause during night  i set ASG size to zero).
It's not running with latest changes/commit , but that one involves only event type detection, nothing have been changed regarding logic.
I have tested  event type detection locally using some sample eventFiles.
So i think is ready to be merged, and maybe tested by more people.
I would like to add more tests, but that can be done later (in instance.go we have a low coverage [66%]).

cristim:
Sounds great, thanks! I'll merge it now

PR #461, opened 2021-05-18
Sequential OnDemand instance replacement through the SQS FIFO queue.

cristim:
# Issue Type

- Feature Pull Request

## Summary

This change implements a few enhancements for the SQS workflow, in order to reduce the amount of time that Spot instances run before being added to the ASG.

The On-demand instance launch events are queued through the SQS FIFO queue and processed one at a time, for each event doing the entire Spot launch and replacement process synchronously before moving to the new event.

This way the ASG capacity is always kept constant and Spot instances are immediately attached to their ASG.

In my initial tests the replacement rate was about 1 instance/minute, but this might be longer if the ASG needs to drain ELB traffic from terminating instances.


cristim (commented):
Circled back on it once again and it looked good to me

PR #462, opened 2021-05-20
Sqs workflow improvements

mello7tre:
guidelines:
* all replacement of ondemand with spot should be handled by sqs
messages and for the same ASG be sequentially
* spot life "outside asg" should be as short as possible
* should be possible to run without using SQSQueue, but ONLY wihout using
event based replacement too (cron only, so that cannot run at the same time two AS execution)
* only one method, swapWithGroupMember, is used to do ondeamnd
replacement

AS = Autospotting
CW = CloudWatch
SQS = Sqs Queue
CRON = Cron Schedule

1) CW ondemand event:
send message to SQS

2) SQS ondemand event:
launch spot replcament - wait for spot running - replace with
swapWithGroupMember

3) CW spot event:
do nothing (it's' needed to avoid intercepting spot launch of point 1)

4) CRON need to launch spot to replace ondemand:
if SQS is enabled we send message to SQS simulating an ondemand
  (ondemand launch)
  message
if SQS is disabled we directly launch spot replacement

5) CRON found spot to attach
if SQS is enabled we send message to SQS (spot launch)
if SQS is disabled we directly call swapWithGroupMember (by
  replaceOnDemandInstanceWithSpot)

6) SQS spot event:
replace with swapWithGroupMember

For point 4 sqs enabled, sending the message to the queue assure a
faster replacement, we do not need to wait for the next cron execution.

To be sure that all replacement for the same ASG (that can change ASG
max size and pause and resume services), are handled in a sequential
way, we use for SQS grouId ONLY the string {Region}-{AsgName}.

# Issue Type

- Feature Pull Request
- Bugfix Pull Request
- Documentation Pull Request

   to it.
   and
   configurations.

### DIFF
--- a/core/action.go
+++ b/core/action.go
@@ -82,3 +82,15 @@ func (ssi swapSpotInstance) run() {
 	spotInstanceID := *ssi.target.spotInstance.InstanceId
 	asg.replaceOnDemandInstanceWithSpot(spotInstanceID)
 }
+
+type sqsSendMessageOnInstanceLaunch struct {
+  target target
+}
+
+func (ssmoil sqsSendMessageOnInstanceLaunch) run() {
+        asg := ssmoil.target.asg
+	onDemandInstanceID := ssmoil.target.onDemandInstance.InstanceId
+	region := ssmoil.target.onDemandInstance.region
+	state := ssmoil.target.onDemandInstance.State.Name
+	region.sqsSendMessageOnInstanceLaunch(&asg.name, onDemandInstanceID, state, "cron-spot-instance-launch")
+}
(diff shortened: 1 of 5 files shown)


cristim:
Why not handle everything in step 5 without SQS?

We could then entirely delete step 6 and have step 5 do the replacement like we anyway do if SQS is disabled.

mello7tre:
because, when cron event start there could be another AS lambda running (triggered by sqs), both acting on the same ASG.

cristim (approved):
Looks pretty good, thanks!

cristim:
oh, I just realized the tests are broken, I'll look into them within the next few days.

mello7tre:
Yes, they was already broken in initial PR. Even gofmt was giving error, but as PR for  `sqs-workflow-improvements` was still a draft, i thought that you was aware of this and was still working on them. So i do not "touched" them.

cristim:
Sounds good, thanks!

I'll take care of the rest next week.

PR #467, opened 2021-09-13
Parameter to disable handling of instance rebalance recommendation events

mello7tre:
# Issue Type

- Feature Pull Request

## Summary
Added a parameter disable_instance_rebalance_recommendation (default to
false) to disable handling of instance rebalance recommendation events.

Often that events are false positive, the spot instance is not
terminated and even if now we do not remove anymore the tag:
launched-for-asg

can happen that, if we detach instance, the rate at which AutoSpotting
cron event terminate unused spot (usually 5min) is greater than the rate at which AWS send IRR events
(usually IRR events are related to the instance type and not
specifically to the instance itself).

What happen is:
AWS send IRR, we deatch spot instance (without decreasing ASG)
ASG launch a new ondemand
if ondemand is in different AZ of the previous detached instance, AS
 cannot use it, so launch a new spot
AWS send IRR for the new spot
and so on ....

AS every cron schedule terminate only one unused spot instance x ASG.
So terminatin rate is lower that new spot + deatch, and detached unused
spot can grow over time.

(whould be usefull to have the parameter overridable by ASG tags, but
for this we need greater change to the code, when we process IRR events
  we do not already know the ASG to which the instance belongs)

   to it.
   and
   configurations.

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -117,6 +117,9 @@ type Config struct {
 
 	// DisableEventBasedInstanceReplacement forces execution in cron mode only
 	DisableEventBasedInstanceReplacement bool
+
+	// DisableInstanceRebalanceRecommendation disable the handling of Instance Rebalance Recommendation events.
+	DisableInstanceRebalanceRecommendation bool
 }
 
 // ParseConfig loads configuration from command line flags, environments variables, and config files.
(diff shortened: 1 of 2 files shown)


mello7tre:
problem arise even because when we handle a new ondemand launch event, we do not check if there is already a spot instance that we can use. We always launch a new spot...

cristim (approved):
Thanks for your contribution!

PR #471, opened 2021-09-16
Feat/ec2 fleet integration

cristim:
# Issue Type

- Feature Pull Request

## Summary

This change switches to using the EC2 Fleet API in [instant](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instant-fleet.html) mode for running Spot instances.

This allows us to support choosing from multiple available [allocation strategies](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html) instead of the lowest-cost strategy previously hardcoded by sorting instance types by price and attempting to run them one after another.

By default we now use the `capacity-optimized-prioritized` allocation strategy. The priority is still based on the lowest cost, we still pass a list ordered by price but with this strategy the ordering is not enforced strongly in case the lowest cost instance types are running low on capacity and have increased risk of interruption, therefore provisioning Spot instances that have lower risk of being interrupted. The cost may not be the very lowest but the interruption rates can sometimes be up to 10x lower than before.

The previous behavior is still available when using the`lowest-cost` allocation strategy.

Another benefit is the use of fewer API calls for launching instances, which will help on large accounts where the previous repeated RunInstances API calls may have contributed to API throttling.

   to it.
   and
   configurations.


cristim:
@mello7tre please try this out when you have some time, this will make Spot instances provisioned by AutoSpotting less prone to be interrupted by tapping into the largest capacity pools.

mello7tre:
Hi @cristim, it's really a massive change!
At the moment i am quite busy, so i cannot tell you when, but i will try to make some tests.

But first i need to have the time to look at the code changes, to understand better the implementation.
Will update you, when i will begin looking at it.

cristim:
Sure, thanks, I really appreciate it! There's no hurry, this isn't final yet.

I tested it a bit and seems to work in my initial tests but I need to test it further and add a few more unit tests before I can confidently merge it. 

The change is smaller than it looks, as it includes a split of instance.go to make it more manageable, so I moved a lot of code around and the tests accompanied to it, which is even bigger.

mello7tre:
Hi @cristim, finally i had some time to look at the code and to study in detail how EC2 Fleet works (never used it).
I still not have tested, on the field, the `fleet version` but i will do it in the next days.

By the way you, have done a great work splitting instance code, i like it a lot, it's very logic; just looking at the names is easy to found the method one is looking for.

cristim:
Thanks, glad you like it. To be honest I was reluctant to split it but when I started it was really too much for my brain so decided to break it down.

I still have to add a bunch more unit tests and test it further but over the last couple of weeks I didn't have time and motivation to work on this.

PR #486, opened 2022-04-12
Fix/price multiplier

mello7tre:
# Issue Type

- Bugfix Pull Request

## Summary

Since 6938027, OnDemandPriceMultiplier were no more taken in account.
The fix is easy, just restore the multiplier when setting instance price (i.price) inside `belongsToEnabledASG`. 

(include a commit that change go.mod to use and updated `ec2-instances-info`)

1. [X] I hereby allow the Copyright holder the rights to distribute this piece of
   to it.
   and
   configurations.

### DIFF
--- a/core/instance_queries.go
+++ b/core/instance_queries.go
@@ -120,7 +120,7 @@ func (i *instance) belongsToEnabledASG() bool {
 			asg.loadLaunchConfiguration()
 			asg.loadLaunchTemplate()
 			i.asg = &asg
-			i.price = i.typeInfo.pricing.onDemand
+			i.price = i.typeInfo.pricing.onDemand / i.region.conf.OnDemandPriceMultiplier * i.asg.config.OnDemandPriceMultiplier
 			log.Printf("%s instace %s belongs to enabled ASG %s", i.region.name,
 				*i.InstanceId, i.asg.name)
 			return true
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/aws/aws-lambda-go v1.26.0
 	github.com/aws/aws-sdk-go v1.40.39
-	github.com/cristim/ec2-instances-info v0.0.0-20210909050335-b239c40fcad0
+	github.com/cristim/ec2-instances-info v0.0.0-20220302182402-4b7a873a84cf
 	github.com/davecgh/go-spew v1.1.1
 	github.com/mattn/goveralls v0.0.9
 	github.com/namsral/flag v0.0.0-20170814194028-67f268f20922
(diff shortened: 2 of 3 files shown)


core/instance_queries.go
```diff
@@ -120,7 +120,7 @@ func (i *instance) belongsToEnabledASG() bool {
 			asg.loadLaunchConfiguration()
 			asg.loadLaunchTemplate()
 			i.asg = &asg
-			i.price = i.typeInfo.pricing.onDemand
+			i.price = i.typeInfo.pricing.onDemand / i.region.conf.OnDemandPriceMultiplier * i.asg.config.OnDemandPriceMultiplier
```
cristim, line 123: Did you test this?

We used to have an issue with this logic(it used to consider the square of the multiplier value instead of the actual value) which I tried to fix a few months ago, but the current fix only considers the parameter from the Lambda environment variables, not from the tag-based ASG overrides.

cristim: right, but what if you also set the value as env var? I suspect it will be squared, because the value comes from both tag and env var and gets multiplied twice.

mello7tre: yes, i suspect that too.
So we should add a condition to ignore `i.region.conf.OnDemandPriceMultiplier` if `i.asg.config.OnDemandPriceMultiplier` is != `1`.
Only one multiplier should be taken in account.
What do you think ?

mello7tre: We calculate region instance ondemand price in method `determineInstanceTypeInformation`.
And we multiply it by the global `OnDemandPriceMultiplier`:
`price.onDemand = it.Pricing[r.name].Linux.OnDemand * cfg.OnDemandPriceMultiplier`

We call `determineInstanceTypeInformation` for both instance launch event and cron event.
So we need to divide by that global value before multiplying by the `OnDemandPriceMultiplier` ASG specific.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/aws/aws-lambda-go v1.26.0
 	github.com/aws/aws-sdk-go v1.40.39
-	github.com/cristim/ec2-instances-info v0.0.0-20210909050335-b239c40fcad0
+	github.com/cristim/ec2-instances-info v0.0.0-20220302182402-4b7a873a84cf
```
cristim, line 8: let's keep this dependency change out of this PR.

go.sum
```diff
@@ -7,6 +7,8 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/cristim/ec2-instances-info v0.0.0-20210909050335-b239c40fcad0 h1:X6u+vwxB2EQwA8+A7jhkYdNK0oE0p0LuuiPbAezZH+g=
 github.com/cristim/ec2-instances-info v0.0.0-20210909050335-b239c40fcad0/go.mod h1:0yCjO4zBzlwWSGh/zGfW2Zq1NX605qCYVBHD1fPXKNs=
+github.com/cristim/ec2-instances-info v0.0.0-20220302182402-4b7a873a84cf h1:/0fth76qMtSWpTbkOLRKHX/k6KgUAK6HMbXArUi7dTc=
+github.com/cristim/ec2-instances-info v0.0.0-20220302182402-4b7a873a84cf/go.mod h1:0yCjO4zBzlwWSGh/zGfW2Zq1NX605qCYVBHD1fPXKNs=
```
cristim, line 11: also here, let's keep this change out of this PR.

core/instance_queries.go
cristim: thanks, I'll look into this over the next few days, I want to address this issue somehow.

cristim (commented):
Taking another look into this, I'll try to sort it out over the next few days.

